PR #65, opened 2016-05-09
Fixed possible crash on exit.

WenhaiLin:


2youyouo2:
cocos-creator/fireball#2901

PR #96, opened 2016-05-12
Solve the crash in simulator if there is no return value of onTouchBegan.

natural-law:


nantas:
cocos-creator/fireball#2557

PR #107, opened 2016-05-16
fixed binary template

2youyouo2:


nantas:
cocos-creator/fireball#3004

PR #116, opened 2016-05-18
fix WebSocket  bug for JSB

knoxHuang:
https://github.com/cocos-creator/fireball/issues/2705


pandamicro:
Fixed https://github.com/cocos-creator/fireball/issues/2838

PR #130, opened 2016-05-19
Fix unzip header url

pandamicro:


pandamicro:
Related to https://github.com/cocos-creator/fireball/issues/3015

PR #132, opened 2016-05-19
IPV6 compatibility

WenhaiLin:


nantas:
cocos-creator/fireball#3470

PR #150, opened 2016-05-27
Fix bindings issues with AssetsManager

pandamicro:


pandamicro:
For https://github.com/cocos-creator/fireball/issues/3111

PR #152, opened 2016-05-30
find Bone after bone get  worldX and worldY bug

knoxHuang:
修复 https://github.com/cocos-creator/fireball/issues/3186


pandamicro:
要提交到这里

https://github.com/EsotericSoftware/spine-runtimes/blob/master/spine-cocos2dx/3/src/spine/SkeletonRenderer.cpp

PR #167, opened 2016-06-13
add JS and JAVA reflection test tips for AppActivity.java

knoxHuang:


templates/js-template-binary/frameworks/runtime-src/proj.android/src/org/cocos2dx/javascript/AppActivity.java
```diff
@@ -33,8 +33,11 @@ of this software and associated documentation files (the "Software"), to deal
 import android.net.wifi.WifiManager;
 import android.os.Bundle;
 import android.view.WindowManager;
+
+// For JS and JAVA reflection test, if you do not need to be deleted
```
pandamicro: you can delete it if it's your own project

PR #226, opened 2016-07-18
fix bug that AudioEngine can not work if file path contains non ascii…

minggo:
… codes

### DIFF
--- a/cocos/audio/apple/AudioPlayer.mm
+++ b/cocos/audio/apple/AudioPlayer.mm
@@ -145,7 +145,7 @@ void AudioPlayer::rotateBufferThread(int offsetFrame)
     ALint bufferProcessed = 0;
     ExtAudioFileRef extRef = nullptr;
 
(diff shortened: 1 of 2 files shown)


cocos/audio/apple/AudioPlayer.mm
```diff
@@ -145,7 +145,7 @@ of this software and associated documentation files (the "Software"), to deal
     ALint bufferProcessed = 0;
     ExtAudioFileRef extRef = nullptr;
 
-    NSString *fileFullPath = [[NSString alloc] initWithCString:_audioCache->_fileFullPath.c_str() encoding:[NSString defaultCStringEncoding]];
+    NSString *fileFullPath = [[NSString alloc] initWithCString:_audioCache->_fileFullPath.c_str() encoding:encoding:NSUTF8StringEncoding];
```
nantas: 这里的修改编译时会报错：
`error: use of undeclared identifier 'encoding'`

PR #253, opened 2016-07-27
Modify the build tools version config in libcocos2dx android studio project.

natural-law:


natural-law:
Related issue: cocos-creator/fireball/issues/3481

PR #254, opened 2016-07-27
revert websocket change to support ipv6 network.

nantas:


nantas:
cocos-creator/fireball#3256

PR #256, opened 2016-07-28
Add image lib ios

dabingnn:
need update external to v3-deps-7


dabingnn:
for issue https://github.com/cocos-creator/fireball/issues/3430

PR #269, opened 2016-08-12
Support show the object group in TiledMap.

natural-law:
@minggo @dumganhar
Please review the PR. Especially for the compatibility. For the details of requirement & solution, please refer to the related issue: cocos-creator/fireball/issues/3120

### DIFF
--- a/cocos/2d/CCFastTMXTiledMap.cpp
+++ b/cocos/2d/CCFastTMXTiledMap.cpp
@@ -160,8 +160,6 @@ void TMXTiledMap::buildWithMapInfo(TMXMapInfo* mapInfo)
     _tileSize = mapInfo->getTileSize();
     _mapOrientation = mapInfo->getOrientation();
 
-    _objectGroups = mapInfo->getObjectGroups();
-
     _properties = mapInfo->getProperties();
 
     _tileProperties = mapInfo->getTileProperties();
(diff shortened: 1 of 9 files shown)


cocos/2d/CCFastTMXTiledMap.cpp
```diff
@@ -217,20 +215,16 @@ TMXLayer * TMXTiledMap::getLayer(const std::string& layerName) const
 
 TMXObjectGroup * TMXTiledMap::getObjectGroup(const std::string& groupName) const
 {
-    CCASSERT(!groupName.empty(), "Invalid group name!");
-    if (groupName.empty()) {
-        return nullptr;
-    }
-
-    if (!_objectGroups.empty())
+    CCASSERT(groupName.size() > 0, "Invalid group name!");
```
dumganhar: Generally, we use empty to check whether a container is empty rather than size > 0.
And if i remember correctly, the `if (groupName.empty()) { return nullptr;}` was added by @WenhaiLin for not causing crash in release mode. Therefore, why to remove this logic?

cocos/2d/CCTMXObjectGroup.h
```diff
@@ -31,24 +31,116 @@ THE SOFTWARE.
...
 
 NS_CC_BEGIN
 
+typedef enum TMXObjectType_ {
+    TMXObjectTypeRect,
+    TMXObjectTypeEllipse,
+    TMXObjectTypePolygon,
+    TMXObjectTypePolyline,
+    TMXObjectTypeImage
+} TMXObjectType;
```
dumganhar: How about using `enum class` which safer. It will look like:
```c++
enum class TMXObjectType
{
    RECT,
    ELLIPSE,
    POLYGON,
    POLYLINE,
    IMAGE
} TMXObjectType;
```

cocos/2d/CCFastTMXTiledMap.cpp
natural-law: In fact, the code are write in cocos2d-x first. Then I apply the changes on the lite repo. So I have no intention to remove the logic. I will add it back.

PR #294, opened 2016-08-31
fix windows systemfont bold

zilongshanren:


zilongshanren:
https://github.com/cocos-creator/fireball/issues/4103

PR #309, opened 2016-09-07
Avoids leak of AudioPlayer if audio file is invalid.

dumganhar:
Invokes GetPrefetchStatus after setPlayState. Make sure there isn't any OpenSLES error.


zilongshanren:
LGTM

PR #325, opened 2016-09-13
Improve the implementation of showing Object Group in TiledMap.

natural-law:
Related issue : cocos-creator/fireball/issues/3120

@dumganhar @pandamicro Please review the PR.

### DIFF
--- a/cocos/2d/CCTMXObjectGroup.h
+++ b/cocos/2d/CCTMXObjectGroup.h
@@ -222,7 +222,7 @@ public:
      *
      * @return The array of the objects.
      */
-    inline const Vector<TMXObject*> getObjects() { return _objects; };
+    inline const Vector<TMXObject*>& getObjects() { return _objects; };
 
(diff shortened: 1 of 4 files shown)


cocos/2d/CCTMXObjectGroup.h
```diff
@@ -147,67 +159,71 @@ class CC_DLL TMXObjectGroup : public Node
...
-
-    /** Gets the array of the objects.
+    
+    /** Gets the array of the objects. 
      *
      * @return The array of the objects.
      */
-    Vector<TMXObject*> getObjects();
-    
+    inline const Vector<TMXObject*> getObjects() { return _objects; };
```
dumganhar: Is there any reason that you don't return a reference or const reference? e.g.
```c++
const Vector<TMXObject*>&
Vector<TMXObject*>&
```

Otherwise, the return vector will be a new vector and will be slower.

PR #367, opened 2016-11-01
Fixing improperly named variable in cc.LabelBMFont.

DavidDeSimone:
There is no width variable defined in cc.LabelBMFont.setWidth. Attempting to use this function will cause a javascript error. This PR will fix that.


zilongshanren:
👍  thanks.

PR #370, opened 2016-11-01
fix editbox can't input on macos sierra

zilongshanren:
1. fix missing some Chinese characters when TTF is encoded with GB2312
2. remove border of EditBox on Mac
3. Fix EditBox can't input text on Mac OS sierra

@pandamicro  This PR depend on https://github.com/cocos-creator/cocos2d-x-lite-external/pull/12

And it should generate a deps-16 external


zilongshanren:
https://github.com/cocos-creator/fireball/issues/4583

PR #371, opened 2016-11-01
Fixing issue where JSScheduleTarget::dump will trigger a bad access

DavidDeSimone:
Due to NULL pointer access.

In 32b8dfd6d8680c82c5b5ba0f52dc88295cea643d @pandamicro fixed an issue where cocos2d_specifics would not compile with COCOS2D_DEBUG=2, however this introduced a logical error that will cause a crash if that flag is set. In the second iteration loop, current will always be NULL and trigger a bad access.


pandamicro:
Oh, sorry, my bad. 

copy paste code is evil...

PR #381, opened 2016-11-08
Fixed EditBox placeholder font not being set correctly for multiline …

DavidDeSimone:
…text fields on iOS.

EditBoxImplIOS::setNativePlaceholderFontColor(const Color4B& color) and EditBoxImplIOS::setNativePlaceholderFont(const char* pFontName, int fontSize) were marked as TODOs. This PR will provided an implementation.


DavidDeSimone:
I'm going to close this PR, as I am going to submit it to mainstream cocos2d-x, which can be integrated into lite at a future date. cocos2d-x PR is here https://github.com/cocos2d/cocos2d-x/pull/16821

PR #409, opened 2016-11-24
reset stencil program when alpha threshold is set to 1

knoxHuang:
https://github.com/cocos-creator/fireball/issues/3961

@minggo 你看看同步的对不对


minggo:
嗯，没问题

PR #419, opened 2016-11-28
Support using specified python by environment 'COCOS_PYTHON_HOME'.

natural-law:
Related issue : cocos-creator/fireball#4803


nantas:
cocos-creator/fireball#4442

PR #437, opened 2016-12-12
iOS project configuration fix.

dumganhar:
AudioDecoder.mm was added to macOS, but it’s missing on iOS, therefore, a linking error will be triggered.


nantas:
update https://github.com/cocos-creator/fireball/issues/4885

PR #442, opened 2016-12-17
Fix crash issues and improve bindings code

pandamicro:
- Fix spine TrackEntry conversion crash
- Fix Wrappers deallocation crash
- Increase JS heap default size to 32 mb
- Improve FinalizeHook implementation


pandamicro:
For https://github.com/cocos-creator/fireball/issues/4675

PR #452, opened 2016-12-21
Sync patch at https://github.com/cocos2d/cocos2d-x/pull/16988

dumganhar:
PR in cocos2d-x: https://github.com/cocos2d/cocos2d-x/pull/16988


dumganhar:
@nantas

PR #473, opened 2017-01-06
Upgrade JS Bindings and AssetsManagerEx

pandamicro:


pandamicro:
For https://github.com/cocos-creator/fireball/issues/4656

PR #507, opened 2017-01-16
fixed FileUtils and downloader bindings in windows

nantas:


pandamicro:

PR #511, opened 2017-01-18
Avoid register too much succeed asset in manifest initialization

pandamicro:


pandamicro:

PR #520, opened 2017-01-24
fix memory leaks of graphics node

zilongshanren:


zilongshanren:
@pandamicro  @jareguo  please review.

PR #521, opened 2017-01-24
Fix AssetsManager issues on x86 platform

pandamicro:


pandamicro:
https://github.com/cocos-creator/fireball/issues/5115

PR #533, opened 2017-02-10
fix label ttf resize_height line wrap

zilongshanren:
fix issue https://github.com/cocos-creator/fireball/issues/5200


cocos/2d/CCLabelTextFormatter.cpp
```diff
@@ -80,14 +80,19 @@ int Label::getFirstCharLen(const std::u16string& utf16Text, int startIndex, int
...
+        || StringUtils::isUnicodeSpace(character)
+        || character == (char16_t)TextFormatter::NewLine)
     {
         return 1;
     }
 
     int len = 1;
     FontLetterDefinition letterDef;
-    auto nextLetterX = 0;
+    if(_fontAtlas->getLetterDefinitionForChar(character, letterDef) == false) {
```
dumganhar: Uses `if (!balabala)` instead of `if (balabala == false)` for boolean type compare.

PR #570, opened 2017-03-15
Issue fixes

pandamicro:


pandamicro:
For https://github.com/cocos-creator/fireball/issues/5408

PR #578, opened 2017-03-22
define rotation property for dragonbone

jareguo:
Re: cocos-creator/fireball#5230

Changes proposed in this pull request:
 * 修复 dragonBones 原生平台未定义 Transform.rotation 属性的问题


natural-law:
LGTM

PR #581, opened 2017-03-29
Adding missing spine property access in jsb for "loop"

DavidDeSimone:
You cannot set the "loop" property from spine jsb. This PR corrects that.


pandamicro:
Thanks ~

PR #603, opened 2017-04-25
Enhance functionalities of AssetsManager

pandamicro:
Fix https://github.com/cocos-creator/fireball/issues/5245 and https://github.com/cocos-creator/fireball/issues/5419


pandamicro:
暂不合并，待添加完整的测试例来验证

PR #624, opened 2017-05-11
add box2d lib dependency to binary win32 template

2youyou2:


nantas:
cocos-creator/fireball#5657

PR #635, opened 2017-05-18
Support multi touch on iOS (the correct way)

pandamicro:


pandamicro:
Fix cocos-creator/fireball#5679

pandamicro:
后续：https://github.com/cocos-creator/fireball/issues/5771

PR #636, opened 2017-05-19
fix pauseAllEventListenersForTarget couldn't work while dispatching touch event.

dumganhar:


dumganhar:
fix the issue: https://github.com/cocos-creator/fireball/issues/5727

PR #639, opened 2017-05-22
Call 'update' method in 'stop/stopAll' to cleanup immediately

dumganhar:


dumganhar:
Fixes issue: https://github.com/cocos-creator/fireball/issues/5783

pandamicro:
验证有效

PR #685, opened 2017-07-05
Uses SocketRocket to implement WebSocket for iOS

dumganhar:
It has the following benefits:

1. The API of SocketRocket is easy for use, it's all asynchronous API.
2. No need to deal with thread communications, it's wrapped inside SocketRocket.
3. Network status change will not make connection closed, it just seems there isn't any status change.
4. Locking screen will trigger an error while libwebsockets or bsd socket will not.
5. Don't depend on openssl library which is a little big.
6. Don't need to use a local CA root certificate (curl ca file is over 2MB) for wss connection
7. Don't trigger signal pipe while closing wss connection
8. Easy to integrate SocketRocket to current WebSocket implementation


pandamicro:
Link https://github.com/cocos-creator/fireball/issues/5994

PR #687, opened 2017-07-05
fixes include in WebSocket-libwebsockets.cpp and updates config.json

dumganhar:


dumganhar:
THIS PR should be merged after the new tag (https://github.com/cocos-creator/cocos2d-x-lite-external/releases) v3-deps-32 is created.

PR #696, opened 2017-07-06
Fix android gen libs issue

pandamicro:


pandamicro:
cocos-creator/fireball#5628

PR #699, opened 2017-07-09
Add back ide-support and fix compilation issues

pandamicro:


pandamicro:
cocos-creator/fireball#5628

PR #711, opened 2017-07-18
Fix new Uint8Array failure on iOS armv7

pandamicro:


pandamicro:
https://github.com/cocos-creator/fireball/issues/6050

PR #729, opened 2017-07-28
Fix restart crash and reduce restart memory leak

pandamicro:


pandamicro:
https://github.com/cocos-creator/fireball/issues/6153

PR #743, opened 2017-08-07
Fix RootedObject on stack being deallocated after ScriptingCore::cleanup

pandamicro:
For https://github.com/cocos-creator/fireball/issues/6168


pandamicro:
For https://github.com/cocos-creator/fireball/issues/6209

PR #744, opened 2017-08-07
Fix anysdk restart issue with old listeners

pandamicro:
For https://github.com/cocos-creator/fireball/issues/6210


pandamicro:
@hylthink 看看有没有疏漏

PR #800, opened 2017-09-06
fixed chinese path bug

huangapple:
fixed chinese path bug


pandamicro:
should fix http://forum.cocos.com/t/creator-1-5-2-windows/50606/2

PR #803, opened 2017-09-07
[jsb2.0] SpiderMonkey debugger works. Adds ScriptEngine::enableDebugger, mainLoopUpdate interfaces.

dumganhar:


dumganhar:
先合并了，有问题再comment。因为要merge一次v1.6

PR #834, opened 2017-09-21
try to fix crash of dragonbones

jareguo:
for https://github.com/cocos-creator/fireball/issues/6387

### DIFF
--- a/cocos/editor-support/dragonbones/animation/AnimationState.h
+++ b/cocos/editor-support/dragonbones/animation/AnimationState.h
@@ -113,7 +113,7 @@ public:
         return _layer;
     }
 
-    inline const std::string& getName() const
+    inline const std::string getName() const
     {
         return _name;
     }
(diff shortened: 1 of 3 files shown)


cocos/editor-support/dragonbones/cocos2dx/CCArmatureDisplay.cpp
```diff
@@ -30,12 +30,12 @@ CCArmatureDisplay::CCArmatureDisplay() :
...
 CCArmatureDisplay::~CCArmatureDisplay() {}
 
+// cleared by armature
 void CCArmatureDisplay::_onClear()
 {
     this->setEventDispatcher(cocos2d::Director::getInstance()->getEventDispatcher());
-
+    clearEventCallback();
     _armature = nullptr;
-    CC_SAFE_RELEASE(_dispatcher);
```
jareguo: _dispatcher should be released in deconstructor

cocos/editor-support/dragonbones/animation/AnimationState.h
```diff
@@ -113,7 +113,7 @@ class AnimationState final : public BaseObject
         return _layer;
     }
 
-    inline const std::string& getName() const
+    inline const std::string getName() const
```
jareguo: a temporary solution

dumganhar: If removes & , then `const` isn't needed.

PR #852, opened 2017-09-26
Image refactor

dumganhar:


cocos/platform/CCImage.h
```diff
@@ -200,14 +194,14 @@ class CC_DLL Image : public Ref
...
     Image& operator=(const Image&);
 
-    /*
-     @brief The same result as with initWithImageFile, but thread safe. It is caused by
-     loadImage() in TextureCache.cpp.
-     @param fullpath  full path of the file.
-     @param imageType the type of image, currently only supporting two types.
-     @return  true if loaded correctly.
+    Image(Image&& rImg);
+    Image& operator=(Image&&);
```
minggo: Why add these functions without implementation?

dumganhar: It's for disable move constructor and operator.
It's a normal way for achieve that.
Since it's private and there isn't any place use it, the following code will trigger errors at compilation time.
```c++
Image a;
Image b = std::move(a); // ERROR
```
It's like what we did in `CC_DISALLOW_COPY_AND_ASSIGN`, but here we also disable `move`.

DavidDeSimone: You should also be able to do Image(Image&& my) = delete; in C++11

minggo: Looks good, it is a pity that VS2013 doesn't support `delete`, or we can use it to make it more explicit.

dumganhar: @DavidDeSimone ,good finding.
I will add `delete` keyword since CocosCreator needs vs2015 at least.

dumganhar: DONE

PR #861, opened 2017-09-26
fix editorBox setText error for engine/issues/1961

knoxHuang:
```js
editBox.setText(text);
int position = text.length();		
editBox.setSelection(position);
```
这样写，会导致部分机型崩溃（错误信息是越界，有些平台可能有最大长度限制），进而修改为下面的方式

```js
editBox.setText(text);
int position = editBox.getText().length();
editBox.setSelection(position);
```

@minggo @dumganhar @jareguo 

https://github.com/cocos-creator/engine/issues/1961


dumganhar:
It's probably `editBox.setText` sets a string longer than its max value.
And setText internally does a clip operation.
Therefore, the string returned by editBox.getText() may shorter than `text.length()`.

PR #866, opened 2017-09-28
use xxtea to decode compiled js files

2youyou2:
https://github.com/cocos-creator/fireball/issues/6289


cocos/scripting/js-bindings/manual/jsb_global.cpp
```diff
@@ -4,27 +4,83 @@
...
+const char* BYTE_CODE_FILE_EXT = ".jsc";
+
+static std::string xxteaKey = "";
+
+void jsb_set_xxtea_key(const std::string& key)
+{
+    xxteaKey = key;
+}
+
+std::string removeFileExt(const std::string& filePath)
```
dumganhar: Adds `static` keyword for `removeFileExt` function.

cocos/scripting/js-bindings/manual/jsb_global.cpp
```diff
@@ -4,27 +4,83 @@
...
 #include "jsb_conversions.hpp"
+#include "xxtea/xxtea.h"
 
 using namespace cocos2d;
 
 se::Object* __jscObj = nullptr;
 se::Object* __ccObj = nullptr;
 se::Object* __jsbObj = nullptr;
 
+const char* BYTE_CODE_FILE_EXT = ".jsc";
```
dumganhar: Adds `static` for BYTE_CODE_FILE_EXT variable.

cocos/scripting/js-bindings/manual/jsb_global.cpp
```diff
@@ -4,27 +4,83 @@
...
+                
+                size_t retLength;
+                unsigned char * data = xxtea_decrypt((unsigned char*)fileData.getBytes(), (uint32_t)fileData.getSize(), (unsigned char*)xxteaKey.c_str(), (uint32_t)xxteaKey.size(), (uint32_t*)&retLength);
+                
+                if (data == nullptr) {
+                    SE_REPORT_ERROR("Can't decrypt code for %s", byteCodePath.c_str());
+                    return "";
+                }
+                
+                return std::string(reinterpret_cast<const char*>(data));
```
dumganhar: Is the data with a `null` terminated letter?
It's better to use `return std::string(reinterpret_cast<const char*>(data), retLength)` to make a safe conversion to string.

cocos/scripting/js-bindings/manual/jsb_global.cpp
```diff
@@ -4,27 +4,83 @@
...
-            Data fileData = FileUtils::getInstance()->getDataFromFile(path);
+            
+            Data fileData;
+            
+#if SCRIPT_ENGINE_TYPE != SCRIPT_ENGINE_SM
+            std::string byteCodePath = removeFileExt(path) + BYTE_CODE_FILE_EXT;
+            if (FileUtils::getInstance()->isFileExist(byteCodePath)) {
+                fileData = FileUtils::getInstance()->getDataFromFile(byteCodePath);
+                
+                *data = xxtea_decrypt((unsigned char*)fileData.getBytes(), (uint32_t)fileData.getSize(), (unsigned char*)xxteaKey.c_str(), (uint32_t)xxteaKey.size(), (uint32_t*)&dataLen);
```
dumganhar: It's will cause memory leak here.
I looked at the code of `xxtea_decrypt`, found that it returns a pointer by `malloc`, therefore, we need to use `free` to release the memory.

`*data = xxtea_decrypt(...)`  uses the `Data` copy assignment which will copy the data rather than take the ownership. so here it will leak.

cocos/scripting/js-bindings/manual/jsb_global.cpp
```diff
@@ -4,27 +4,83 @@
...
         delegate.onGetStringFromFile = [](const std::string& path) -> std::string{
             assert(!path.empty());
+            
+#if SCRIPT_ENGINE_TYPE != SCRIPT_ENGINE_SM
+            std::string byteCodePath = removeFileExt(path) + BYTE_CODE_FILE_EXT;
+            if (FileUtils::getInstance()->isFileExist(byteCodePath)) {
+                Data fileData = FileUtils::getInstance()->getDataFromFile(byteCodePath);
+                
+                size_t retLength;
+                unsigned char * data = xxtea_decrypt((unsigned char*)fileData.getBytes(), (uint32_t)fileData.getSize(), (unsigned char*)xxteaKey.c_str(), (uint32_t)xxteaKey.size(), (uint32_t*)&retLength);
```
dumganhar: Same here.

PR #870, opened 2017-09-29
remove copy_resources from android-studio build-cfg.json

2youyou2:
android-studio copy resources in `proj.android-studio/app/build.gradle`, so do not need to copy resources in build-cfg.json


dumganhar:
How about some people are still using `ant` to build apk? This will break that.

2youyou2:
proj.android and proj.android-studio use different build-cfg.json

PR #872, opened 2017-09-30
Adds docs for JSB2.0

dumganhar:


DavidDeSimone:
@dumganhar first off, thank you for all your great work on JSB 2.0. Are there any plans to have these docs in English as well?

dumganhar:
@DavidDeSimone, English version is DONE at https://github.com/cocos-creator/cocos2d-x-lite/blob/v1.7/cocos/scripting/js-bindings/docs/JSB2.0-learning-en.md

PR #873, opened 2017-10-06
Remove scheduleCallbackForTarget with selector usage

pandamicro:
Re: cocos-creator/fireball#6418


pandamicro:
selector usage will cause memory leak for scheduleOnce, after scheduler stop the scheduleOnce task, it won't release the SchedulerWrapper for it

PR #879, opened 2017-10-09
Polish jsbinding code and debugger & profiler work

dumganhar:
This PR depends on https://github.com/cocos-creator/bindings-generator/pull/18 and https://github.com/cocos-creator/cocos2d-console/pull/48

* Adds jsb_set_extend_property for easily setting extend property. 
* Adds jsb_init_file_operation_delegate and jsb_enable_debugger for easily use in AppDelegate.cpp.
* Safe check in dragonbones cleanup.
* se::Object::setProperty should return boolean.
* se::ScriptEngine::enableDebugger takes serverAddr and port arguments.
* Moves cocos2d-x node:onEnter/onExit binding code from se::ScriptEngine to ScriptingCore.
* Updates config.hpp, uses COCOS2D_DEBUG for controlling debugger.
* Print JS engine version
* Debugger and profiler work
* Updates docs for debugging and profiling.


cocos/scripting/js-bindings/jswrapper/jsc/ScriptEngine.hpp
```diff
@@ -225,18 +232,6 @@ namespace se {
...
-        {
-            ENTER,
-            EXIT,
-            ENTER_TRANSITION_DID_FINISH,
-            EXIT_TRANSITION_DID_START,
-            CLEANUP
-        };
-        bool _onReceiveNodeEvent(void* node, NodeEventType type);
-        using NodeEventListener = bool(*)(void*, NodeEventType);
-        bool _setNodeEventListener(NodeEventListener listener);
```
dumganhar: jswrapper should be clean which means it should not depend on any concepts in Cocos2d-x. 
So, node event callback should be moved.

cocos/scripting/js-bindings/jswrapper/sm/Object.cpp
```diff
@@ -176,13 +176,13 @@ namespace se {
...
 
-    void Object::setProperty(const char* name, const Value& v)
+    bool Object::setProperty(const char* name, const Value& v)
     {
         JS::RootedObject object(__cx, _getJSObject());
 
         JS::RootedValue value(__cx);
         internal::seToJsValue(__cx, v, &value);
-        JS_SetProperty(__cx, object, name, value);
+        return JS_SetProperty(__cx, object, name, value);
```
dumganhar: se::Object::setProperty should return boolean value to indicate whether the invocation is successful.

cocos/scripting/js-bindings/manual/jsb_global.cpp
```diff
@@ -103,7 +105,59 @@ bool jsb_run_script(const std::string& filePath)
...
+            se::ScriptEngine::getInstance()->mainLoopUpdate();
+        }
+    };
+    static SimpleRunLoop runLoop;
+    Director::getInstance()->getScheduler()->scheduleUpdate(&runLoop, 0, false);
+    return true;
+}
+
+bool jsb_set_extend_property(const char* ns, const char* clsName)
+{
```
dumganhar: A wrapper for easily setting xxx.extend = cc.Class.extend.
Don't use se::ScriptEngine::evalString since it will shown in `Browser Debugger` with `(no filename)`. And it's useless.

PR #935, opened 2017-10-26
Auto binds RenderTexture::saveToFile and fix js memory leak while saveToFile.

dumganhar:
Fixes https://github.com/cocos-creator/fireball/issues/6527

The callback _saveFileCallback has to be set to nullptr to release se::Value the std::function captures.

Test code:
```js
             let dirpath = jsb.fileUtils.getWritablePath() + 'ScreenShoot/';
             if( !jsb.fileUtils.isDirectoryExist(dirpath)){
                 jsb.fileUtils.createDirectory(dirpath);
             }
             let name = 'ScreenShoot-' + (new Date()).valueOf() + '.png';
             let filepath = dirpath + name;
             let size = cc.winSize;
             let rt = cc.RenderTexture.create(size.width, size.height);
             cc.director.getScene()._sgNode.addChild(rt);
             rt.setVisible(false);
             rt.begin();
             cc.director.getScene()._sgNode.visit();
             rt.end();
             var func = function(path) {
                 console.log(path);
             };

             rt.saveToFile('ScreenShoot/' + name, cc.ImageFormat.PNG, true, function() {
                           cc.log('save succ');
                           rt.removeFromParent();
                           if (func) {
                           func(filepath);
                           }
                           // memory leak if lambda isn't set to null or `rt` isn't set to null.
                           // rt = null; // Adds rt = null to avoid memory leak.
                           });
```

### DIFF
--- a/cocos/2d/CCRenderTexture.cpp
+++ b/cocos/2d/CCRenderTexture.cpp
@@ -746,6 +746,7 @@ void RenderTexture::draw(Renderer *renderer, const Mat4 &transform, uint32_t fla
 
 void RenderTexture::cleanup()
 {
+    Node::cleanup();
     _saveFileCallback = nullptr;
 }
 
--- a/cocos/2d/CCRenderTexture.cpp
+++ b/cocos/2d/CCRenderTexture.cpp
@@ -744,6 +744,11 @@ void RenderTexture::draw(Renderer *renderer, const Mat4 &transform, uint32_t fla
     }
 }
 
+void RenderTexture::cleanup()
+{
+    _saveFileCallback = nullptr;
+}
+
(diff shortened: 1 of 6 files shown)


cocos/2d/CCRenderTexture.cpp
```diff
@@ -744,6 +744,12 @@ void RenderTexture::draw(Renderer *renderer, const Mat4 &transform, uint32_t fla
     }
 }
 
+void RenderTexture::cleanup()
+{
+    Node::cleanup();
+    _saveFileCallback = nullptr;
```
dumganhar: Sets `std::function` to nullptr to avoid lambda captures js function.

PR #952, opened 2017-11-02
Bug fix for jsb.callStaticMethod for android. Don't return the value JavaScriptJavaBrider::convertReturnValue returns for binding method.

dumganhar:


cocos/scripting/js-bindings/manual/JavaScriptJavaBridge.cpp
```diff
@@ -483,8 +482,8 @@ static bool JavaScriptJavaBridge_callStaticMethod(se::State& s)
                 SE_REPORT_ERROR("call result code: %d", call.getErrorCode());
                 return false;
             }
-            ok = JavaScriptJavaBridge::convertReturnValue(call.getReturnValue(), call.getReturnValueType(), &s.rval());
-            return ok;
+            JavaScriptJavaBridge::convertReturnValue(call.getReturnValue(), call.getReturnValueType(), &s.rval());
+            return true;
```
dumganhar: Have to return true here.

PR #954, opened 2017-11-02
Empty implementation for the platforms that don't support Cocos Analytics.

dumganhar:


dumganhar:
github出bug了，没法指定reviewer,
手动 @jareguo

PR #966, opened 2017-11-07
add setEnable to CameraNode

2youyou2:
https://github.com/cocos-creator/fireball/issues/6588


cocos/editor-support/creator/CCCameraNode.cpp
```diff
@@ -34,7 +33,16 @@ namespace creator {
             }
             _commands.clear();
         }
-        _cameraInstance = nullptr;
+    }
+    
+    void CameraNode::setEnable(bool enable)
```
2youyou2: 由 js 层控制使用哪个 camera

PR #979, opened 2017-11-10
Uses shorter module name (cocos2dx_internal_static -> cis) to avoid compilation errors on Windows system which has a limitation of 260 bytes for build path.

dumganhar:
Yep, probably, we could define `buildDir` variable in build.gradle, but let's investigate later.


jareguo:
让我笑一会

WenhaiLin:
恭喜码神成功腾出22个字符空间 😈

jareguo:
fixed https://github.com/cocos-creator/fireball/issues/6621

PR #1006, opened 2017-11-22
Delete old version files during hot update

pandamicro:


jareguo:
for https://github.com/cocos-creator/fireball/issues/6656

PR #1011, opened 2017-11-23
Some Android Simulators don't support SSE instruction, so disable it for x86 arch.

dumganhar:
Fixed https://github.com/cocos-creator/fireball/issues/6666

### DIFF
--- a/templates/js-template-binary/frameworks/runtime-src/proj.android-studio/app/jni/Application.mk
+++ b/templates/js-template-binary/frameworks/runtime-src/proj.android-studio/app/jni/Application.mk
@@ -26,3 +26,5 @@ else
   APP_OPTIM := release
 endif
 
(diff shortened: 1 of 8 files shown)


templates/js-template-binary/frameworks/runtime-src/proj.android-studio/app/jni/Application.mk
```diff
@@ -26,3 +26,5 @@ else
   APP_OPTIM := release
 endif
 
+# Some Android Simulators don't support SSE instruction, so disable it for x86 arch.
+APP_CPPFLAGS += -U__SSE__
```
dumganhar: Can't use 
```
ifeq ($(TARGET_ARCH_ABI),x86)
APP_CPPFLAGS += -U__SSE__
endif
```
since TARGET_ARCH_ABI isn't available in Application.mk

PR #1036, opened 2017-12-19
Corrects the wrong logic of using delete to remove memory allocated by malloc.

dumganhar:
解决 GraphicsNode 中用 delete 去释放 malloc 申请的内存。


2youyou2:
LGTM

PR #1048, opened 2017-12-31
refactor dragonbones asset managment

jareguo:
for https://github.com/cocos-creator/fireball/issues/6749

allow dragonbone parse reloaded TextureAtlasData directly


2youyou2:
LGTM

PR #1075, opened 2018-01-18
Bug fix of releasing JS object logic in CPP controls JS object's cycle mode (spine, dragonbones, box2d).

dumganhar:
解决 spine, box2d, dragonbones 绑定中可能触发 se::Object::setPrivateData 中的断言的问题。


dumganhar:
creator docs: https://github.com/cocos-creator/creator-docs/pull/244

PR #1112, opened 2018-02-11
mac main loop work

minggo:


stevetranby:
Are these being ported from the main repo?
What % of commits to this repo are new?
Is this repo planned as v4 of the c++ engine, or is this just for the CocosCreator native runtime?

minggo:
It is just for CocosCreator, but v4 will share some codes with it, such as new renderer.

PR #1178, opened 2018-04-02
Fix local reference table overflow error problem.

wzpan:
In our app, local reference table overflow errors occur frequently. 

After investigating, I found that such problem is caused by the `NewStringUTF()` function calling, which does not follows a `DeletLocalRef()` operation.

This pull request fixes it. It works well in our app and no more local reference table error occurs again.

### DIFF
--- a/cocos/scripting/js-bindings/manual/JavaScriptJavaBridge.cpp
+++ b/cocos/scripting/js-bindings/manual/JavaScriptJavaBridge.cpp
@@ -508,10 +508,7 @@ static bool JavaScriptJavaBridge_callStaticMethod(se::State& s)
         {
             int count = argc - 3;
             jvalue* jargs = new jvalue[count];
-            int *toRelease = new int[count];
-            for (int i = 0; i < count; ++i){
-                toRelease[i] = 0;
-            }
+            std::vector<jobject> toReleaseObjects;
             for (int i = 0; i < count; ++i)
             {
                 int index = i + 3;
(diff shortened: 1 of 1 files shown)


dumganhar:
@wzpan  Good catch.
But I have some comments.

cocos/scripting/js-bindings/manual/JavaScriptJavaBridge.cpp
```diff
@@ -508,6 +508,10 @@ static bool JavaScriptJavaBridge_callStaticMethod(se::State& s)
         {
             int count = argc - 3;
             jvalue* jargs = new jvalue[count];
+            int *toRelease = new int[count];
```
dumganhar: Where to delete[] toRelease array?
Would using `std::vector` be better?

wzpan: Yes you are right. I Will modify a bit to meet this comment.

cocos/scripting/js-bindings/manual/JavaScriptJavaBridge.cpp
```diff
@@ -508,6 +508,10 @@ static bool JavaScriptJavaBridge_callStaticMethod(se::State& s)
...
+            for (int i = 0; i < count; ++i){
+                toRelease[i] = 0;
+            }
```
dumganhar: NO need if using `std::vector`

cocos/scripting/js-bindings/manual/JavaScriptJavaBridge.cpp
```diff
@@ -544,10 +548,16 @@ static bool JavaScriptJavaBridge_callStaticMethod(se::State& s)
                         std::string str;
                         seval_to_std_string(args[index], &str);
                         jargs[i].l = call.getEnv()->NewStringUTF(str.c_str());
+                        toRelease[i] = 1;
```
dumganhar: toReleaseObjects.push_back(jargs[i].l);

wzpan:
@dumganhar I've modified the patch according to your review comment. Please review it again.

dumganhar:
@wzpan , merged ! Thank you.

PR #1198, opened 2018-04-18
Throw args away rather than raising an error.

GAMEKON:
Throw args away rather than raising an error.


dumganhar:
Thanks. Merged.

PR #1211, opened 2018-04-24
add android instant app support for cocos

wuzhiming:
add android instant app support for cocos


dumganhar:
Why to commit gradle-wrapper.jar ?

PR #1222, opened 2018-04-27
fix downloader not continue for android

drelaptop:
https://github.com/cocos2d/cocos2d-x/pull/17761

download not continue occurs when the values:
`countOfMaxProcessingTasks(jsb.Downloader)` <= `cc.macro.DOWNLOAD_MAX_CONCURRENT`

this PR fix it

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxDownloader.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxDownloader.java
@@ -87,6 +87,12 @@ class DataTaskHandler extends BinaryHttpResponseHandler {
         LogD("onSuccess(i:" + i + " headers:" + headers);
         _downloader.onFinish(_id, 0, null, binaryData);
     }
+    
+    @Override
+    public void onFinish() {
+        // onFinish called after onSuccess/onFailure
+        _downloader.runNextTaskIfExists();
+    }
 }
 
 class HeadTaskHandler extends AsyncHttpResponseHandler {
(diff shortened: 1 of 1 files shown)


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxDownloader.java
```diff
@@ -327,9 +339,6 @@ public void run() {
                     catch (URISyntaxException e) {
                         break;
                     }
-                    catch (NullPointerException e) {
-                        break;
```
dumganhar: Why not to catch NullPointerException?

drelaptop: this is sync the changes from cocos2d-x repo, I don't know this clearly yet. and I don't think it's a good behavior to do nothing after catch a Exception. we should print warn-level log at least.

dumganhar: Yep, log output is better. But if we don't know the exact reason of removing `catch Null Pointer Exception` line, let's revert it at least. 
I agree that adding more detailed log output in NullPointerException catching block.

drelaptop: thanks for advice, I will research the reason

drelaptop: See the comments, `NullPointerException` will throw  when the URL is null
```java
     * @throws  NullPointerException
     *          If {@code str} is {@code null}
    public URI(String str) throws URISyntaxException {
        new Parser(str).parse(false);
    }
```
in  other side, before call those java code, the native code checked the URL
```C++
    std::shared_ptr<const DownloadTask> Downloader::createDownloadFileTask(const std::string& srcUrl,
...
    {
        do
        {
            ...
            if (0 == srcUrl.length() || 0 == storagePath.length())
            {
                if (onTaskError)
                {
                    onTaskError(*task, DownloadTask::ERROR_INVALID_PARAMS, 0, "URL or storage path is empty.");
                }
                task.reset();
                break;
            }
...
```
So, the `NullPointerException` will never occur.

PR #1234, opened 2018-05-10
fix bmfont kerning parsing bug

umbrellaPP:
Re: cocos-creator/fireball#7136

http://forum.cocos.com/t/bmfont-native-1-8-2/58466

### DIFF
--- a/cocos/2d/CCFontFNT.cpp
+++ b/cocos/2d/CCFontFNT.cpp
@@ -801,8 +801,8 @@ int * FontFNT::getHorizontalKerningForTextUTF16(const std::u16string& text, int 
 
     for (int c = 0; c < outNumLetters; ++c)
     {
-        if (c < (outNumLetters-1))
-            sizes[c] = getHorizontalKerningForChars(text[c], text[c+1]);
+        if (c > 0)
+            sizes[c] = getHorizontalKerningForChars(text[c-1], text[c]);
         else
             sizes[c] = 0;
     }


cocos/2d/CCLabelTextFormatter.cpp
```diff
@@ -220,7 +220,7 @@ bool Label::multilineTextWrap(const std::function<int(const std::u16string&, int
             if (nextChangeSize)
             {
                 if (_horizontalKernings && letterIndex < textLen - 1)
-                    nextLetterX += _horizontalKernings[letterIndex + 1];
+                    nextLetterX += _horizontalKernings[letterIndex];
```
drelaptop: I don't think it's a good bug fix for BMFont issue, for this function `Label::multilineTextWrap` is a common logic for all fonts, not only BMFont, but also TTF font. So this change the TTF font wrap logic for a BMFont issue only.

you should change the BMFont `_horizontalKernings ` logic only, so follow the logic line
```
nextLetterX += _horizontalKernings[letterIndex];
->  _horizontalKernings
-> Label::computeHorizontalKernings
-> _fontAtlas->getFont()->getHorizontalKerningForTextUTF32
-> FontFNT::getHorizontalKerningForTextUTF32
```
this is cocos2d-x logic, I think similar logic existed cocos2d-x-lite. please change the BMFont logic only, secondly make sure you can reproduce  that issue, and solve this issue after your changes, meanwhile not bad effect on all label test cases.

drelaptop:
@dumganhar please merge

PR #1245, opened 2018-05-17
Downsample

minggo:


cocos/base/CCGLUtils.cpp
```diff
@@ -24,151 +24,204 @@
...
-    {
-        __currentBoundTexture[i] = -1;
-    };
+    __currentVertexBuffer = -1;
+    __currentIndexBuffer = -1;
+    __currentVertexArray = -1;
+    
+    __enabledVertexAttribArrayFlag = 0;
+    for (int i = 0; i < MAX_ATTRIBUTE_UNIT; ++i)
+        __enabledVertexAttribArrayInfo[i] = VertexAttributePointerInfo();
```
dumganhar: It's unneeded to initialize __enabledVertexAttribArrayInfo array since all elements in it will be initialized by VertexAttributePointerInfo constructor.

cocos/base/CCGLUtils.cpp
```diff
@@ -24,151 +24,204 @@
...
-//        {
-//            printf("glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, %u) isn't needed!\n", buffer);
-//        }
     }
     else
     {
         glBindBuffer(target, buffer);
     }
+
+    CHECK_GL_ERROR_DEBUG();
```
dumganhar: It's better to CHECK GL Error outside `ccBindBuffer` because we just wanna simulate `gl` functions which will not do this.

minggo: Yep, agree.

cocos/platform/ios/CCApplication-ios.mm
```diff
@@ -177,13 +191,17 @@ -(void) doCaller: (id) sender
...
+Application::Application(const std::string& name, int width, int height)
 {
     Application::_instance = this;
     _scheduler = new Scheduler();
 
     createView(name);
     
+    // Shoule create _renderTexture here, then we don't have to consider how to
+    // revert the GL states. It is difficult to revert the states.
+    _renderTexture = new RenderTexture(width, height, 2);
```
dumganhar: Perhaps, `RenderTextureScreen` is a better name since current `RenderTexture` isn't a command module.
It's only for devicePixelRatio functionality.

PR #1249, opened 2018-05-18
Move logic of flipY and premultiplyAlpha from jsb_opengl_manual.cpp to CCGLUtils.cpp.

dumganhar:
Depends on https://github.com/cocos-creator/engine/pull/2700.

And some fixes for native  renderer.


minggo:
可以合了

PR #1307, opened 2018-06-07
default template works on android studio

drelaptop:


templates/js-template-default/frameworks/runtime-src/proj.android-studio/gradle.properties
```diff
@@ -17,10 +17,26 @@
...
+# To build for multiple architexture, use the `:` between them
+# Example - PROP_APP_ABI=armeabi-v7a:arm64-v8a:x86
 PROP_APP_ABI=armeabi-v7a
 
-RELEASE_STORE_FILE=
-RELEASE_STORE_PASSWORD=
-RELEASE_KEY_ALIAS=
-RELEASE_KEY_PASSWORD=
+# uncomment it and fill in sign information for release mode
+RELEASE_STORE_FILE=/Users/laptop/code-github/fireball/editor/static/build-templates/native/debug.keystore
```
dumganhar: ?? Hardcode for store file?

templates/js-template-default/frameworks/runtime-src/proj.android-studio/app/build.gradle
```diff
@@ -19,8 +19,7 @@ android {
                     // skip the NDK Build step if PROP_NDK_MODE is none
                     targets 'cocos2djs'
                     arguments 'NDK_TOOLCHAIN_VERSION=4.9'
-                    arguments 'APP_PLATFORM=android-'+PROP_TARGET_SDK_VERSION
```
dumganhar: APP_PLATFORM isn't needed?

cocos/Android.mk
```diff
@@ -71,7 +71,7 @@ renderer/renderer/Light.cpp \
 renderer/renderer/Model.cpp \
 renderer/renderer/Pass.cpp \
 renderer/renderer/ProgramLib.cpp \
-renderer/renderer//Scene.cpp \
+renderer/renderer/Scene.cpp \
```
dumganhar: Good catch.

templates/js-template-default/frameworks/runtime-src/proj.android-studio/gradle.properties
drelaptop: my god, it's a mistake. I will re-check it tomorrow.

templates/js-template-default/frameworks/runtime-src/proj.android-studio/app/build.gradle
drelaptop: yes, it always equal mini sdk version, refer to below. discuss at https://github.com/cocos2d/cocos2d-x/pull/18646

> @newnon Sounds reasonable, but android SDK will set it on build automatic? how to prove that, or is there some official docs said that?

> sure https://developer.android.com/ndk/guides/application_mk.html
>APP_PLATFORM
This variable contains the minimum Android platform version you want to support. For example, a value of android-15 specifies that your library uses APIs that are not available below Android 4.0.3 (API level 15) and can't be used on devices running a lower platform version. For a complete list of platform names and corresponding Android system images, see Android NDK Native APIs.
>Instead of changing this flag directly, you should set the minSdkVersion property in the defaultConfig or productFlavors blocks of your module-level build.gradle file. This makes sure your library is used only by apps installed on devices running an adequate version of Android. The ndk-build toolchain uses the following logic to choose the minimum platform version for your library based the ABI you're building and the minSdkVersion you specify:
>If there exists a platform version for the ABI equal to minSdkVersion, ndk-build uses that version.
Otherwise, if there exists platform versions lower than minSdkVersion for the ABI, ndk-build uses the highest of those platform versions. This is a reasonable choice because a missing platform version typically means that there were no changes to the native platform APIs since the previous available version.
Otherwise, ndk-build uses the next available platform version higher than minSdkVersion.

templates/js-template-default/frameworks/runtime-src/proj.android-studio/app/jni/Application.mk
```diff
@@ -10,7 +10,7 @@ APP_LDFLAGS := -latomic
 APP_SHORT_COMMANDS := true
 
 USE_ARM_MODE := 1
-USE_ANY_SDK := 1
+#USE_ANY_SDK := 1
```
drelaptop: should I remove all anysdk related code in Android.mk?

templates/js-template-default/frameworks/runtime-src/proj.android-studio/gradle.properties
```diff
@@ -17,10 +17,26 @@
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
 
-PROP_TARGET_SDK_VERSION=22
+# Android SDK version that will be used as the compile project
+PROP_COMPILE_SDK_VERSION=27
+
+# Android SDK version that will be used as the earliest version of android this application can run on
+PROP_MIN_SDK_VERSION=16
```
drelaptop: what min sdk version and app platform version you used to build external android libs? set it 16, compile success, 14 ,compile failed.

`PROP_MIN_SDK_VERSION=16` means min sdk version and app platform version both are 16, you can search External Native Build log to verify this.

cocos/Android.mk
```diff
@@ -165,7 +165,7 @@ LOCAL_WHOLE_STATIC_LIBRARIES := cocos2dxandroid_static
 LOCAL_WHOLE_STATIC_LIBRARIES += cpufeatures
 
 # define the macro to compile through support/zip_support/ioapi.c
-LOCAL_CFLAGS := -DUSE_FILE32API -fexceptions -DHAVE_INSPECTOR
```
drelaptop: `HAVE_INSPECTOR` have defined in `config.hpp`, like this
```
#if defined(COCOS2D_DEBUG) && COCOS2D_DEBUG > 0
#define SE_ENABLE_INSPECTOR 1
#define SE_DEBUG 2
#define HAVE_INSPECTOR 1
#else
#define SE_ENABLE_INSPECTOR 0
#define SE_DEBUG 0
#define HAVE_INSPECTOR 0
#endif
```
it cause many re-defined errors when building native code.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxGLSurfaceView.java
```diff
@@ -128,7 +126,6 @@ public void onPause() {
             @Override
             public void run() {
                 Cocos2dxGLSurfaceView.this.mCocos2dxRenderer.handleOnPause();
-                CAAgent.onPause(Cocos2dxGLSurfaceView.this.getContext());
```
drelaptop: remove all AnySDK related code, or you will see error logs like
```
06-08 11:24:12.552 12641-12681/org.cocos2d.helloworld E/CocosAnalytics: ==> checkAndUploadLog, database is not open!
06-08 11:24:42.554 12641-12681/org.cocos2d.helloworld E/CocosAnalytics: ==> Please init sdk first!
```

templates/js-template-default/frameworks/runtime-src/proj.android-studio/app/jni/hellojavascript/main.cpp
```diff
@@ -28,36 +28,14 @@
...
-#endif
-    return pAppDelegate;
-}
-
-extern "C"
-{
-	void Java_org_cocos2dx_javascript_SDKWrapper_nativeLoadAllPlugins(JNIEnv*  env, jobject thiz)
-	{
-#if PACKAGE_AS
-    	SDKManager::getInstance()->loadAllPlugins();
```
drelaptop: since we don't needed AnySDK anymore, this should be removed too. __main.cpp__ looks so clear with AnySDk code removed

templates/js-template-default/frameworks/runtime-src/proj.android-studio/app/jni/Application.mk
drelaptop: have removed

drelaptop:
Tested by hello_world project, wroks well with those Android configs
```
PROP_COMPILE_SDK_VERSION=27
PROP_MIN_SDK_VERSION=16
PROP_TARGET_SDK_VERSION=27
PROP_BUILD_TOOLS_VERSION=27.0.3
PROP_APP_ABI=armeabi-v7a:arm64-v8a:x86
```
Device, Android 8.0, API 26

PR #1310, opened 2018-06-11
link template works on mac, ios, android studio 

drelaptop:
just like those default template do, only with little difference.

default template on mac, #1300 
default template on ios, #1304 
default template on android studio, #1307


drelaptop:
please review and merge @minggo

templates/js-template-link/frameworks/runtime-src/Classes/AppDelegate.cpp
```diff
@@ -44,7 +43,7 @@
 
 USING_NS_CC;
 
-AppDelegate::AppDelegate() : Application("js-test")
+AppDelegate::AppDelegate(int width, int height) : Application("Cocos Runtime", width, height)
```
minggo: `Cocos Runtime` is not suitable here.

drelaptop: I think so, hard coded the application name is a bad choice, might we can add a macro
```
// modify it to change application name
#define COCOS_APPLICATION_NAME "Cocos Game"
```
for further, we can modify it by cocos console, for example change default value to `HelloJavaScript`, change it to `hellowrold` when the identifier is `org.cocos2d.helloworld`

drelaptop:
changed "Cocos Runtime" into "Cocos Game", I think "Cocos Game" is a better app name than "js-tests". please review @minggo

minggo:
Yep, `Cocos Game` is good.

PR #1313, opened 2018-06-13
gen-simulator works on mac, but not full features

drelaptop:
## Attention

View control features are lost, for those reason.

1. needed native feature disappear at native runtime, for example
```
auto displayStats = Director::getInstance()->isDisplayStats(); // display FPS or not
Director::getInstance()->getOpenGLView()->setFrameZoomFactor(scale); // scale
```
2. those codes of Simulator are confusing, for example
what's the meaning of adding `PlayerEditBoxServiceProtocol` class into simulator

## Tests

image


tools/simulator/frameworks/runtime-src/proj.ios_mac/mac/SimulatorApp.mm
```diff
@@ -381,27 +380,28 @@ - (void) createWindowAndGLView
...
-    [self setZoom:_project.getFrameScale()];
+    // [self setZoom:_project.getFrameScale()];
     if (pos.x != 0 && pos.y != 0)
     {
         [_window setFrameOrigin:NSMakePoint(pos.x, pos.y)];
     }
 
 #if (PLAYER_SUPPORT_DROP > 0)
-    glfwSetDropCallback(eglView->getWindow(), glfwDropFunc);
+    glfwSetDropCallback((GLFWwindow*)glfwWindow, glfwDropFunc);
```
drelaptop: `glfwSetDropCallback ` is used to monitor drop event, for example drop a file into cocos simulator window.
> This function sets the file drop callback of the specified window, which is called when one or more dragged files are dropped on the window.

this feature wasn't used, even cocos creator 1.9.x, might need to be removed later. and much more code have no related with simulator.

tools/simulator/frameworks/runtime-src/proj.ios_mac/mac/SimulatorApp.mm
```diff
@@ -446,19 +446,18 @@ - (void) startup
...
     FileUtils::getInstance()->addSearchPath(resourcePath.UTF8String);
 
-    // app
-    _app = new AppDelegate();
-
-    [self setupUI];
-    [self adjustEditMenuIndex];
+    // not show the custom menu items
+    // [self setupUI];
+    // [self adjustEditMenuIndex];
```
drelaptop: custom menu items needed native feature disappear at native runtime. such as, set window scale, display fps or not.

So I comment the `setupUI` code, now the simulator works well without this feature

tools/simulator/libsimulator/lib/runtime/ConnectWaitLayer.cpp
```diff
@@ -24,7 +24,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 ****************************************************************************/
 
-#include "ConnectWaitLayer.h"
+//#include "ConnectWaitLayer.h"
```
drelaptop: comment other code, those are simulator no related, and cause compile error

drelaptop:
@minggo please review

tools/simulator/frameworks/runtime-src/proj.ios_mac/mac/SimulatorApp.mm
minggo: I think these features are needed, and should be implemented in JS later.

tools/simulator/frameworks/runtime-src/proj.ios_mac/mac/SimulatorApp.mm
minggo: Sounds reasonable, it is better to add TODO in the codes to avoid forgetting it.

PR #1319, opened 2018-06-19
[adjust] Provide a interface for audio interrupted in Cocos2dxAudioFocusManager

wdzhangchukong:


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxAudioFocusManager.java
```diff
@@ -114,13 +154,6 @@ static void unregisterAudioFocusListener(Context context) {
         } else {
             Log.e(TAG, "abandonAudioFocus failed!");
         }
-
-        Cocos2dxHelper.runOnGLThread(new Runnable() {
-            @Override
-            public void run() {
-                nativeOnAudioFocusChange(AUDIOFOCUS_GAIN);
-            }
-        });
```
minggo: Why remove it?

wdzhangchukong: UnregisterAudioFocusListener will be invoked when app on pause, and audio focus will lose, so I think it doesn't need to gain the audio focus.

PR #1320, opened 2018-06-19
use gzip to decompress js content

2youyou2:
sync develop branch


pandamicro:
For https://github.com/cocos-creator/fireball/issues/7501

PR #1330, opened 2018-06-26
[Not Finished] win32 Text Rendering Context2D works 

drelaptop:
it works for simple single-line label now, and is going on


dumganhar:
Cool

PR #1333, opened 2018-06-27
correct label render contex2d

drelaptop:
go on after #1330 did

for issue, https://github.com/cocos-creator/fireball/issues/7630


drelaptop:
## win32 label contex2d finished with little issue

1. outline have no effect
2. text lost __only__ on __RichTextImage__ test case in examplecases test project

@minggo please merge

## example

text-render

PR #1334, opened 2018-06-27
remove gen-libs

jareguo:
for https://github.com/cocos-creator/fireball/issues/7630


drelaptop:
I think this PR should been merged @minggo

PR #1338, opened 2018-06-28
correct copy jsb-adapter folder when build

drelaptop:
go on #1336 

for issues cocos-creator/fireball#7630


drelaptop:
win32/ Android/ iOS/ macOS ，模板工程，验证完成。请合并 @minggo

PR #1370, opened 2018-07-05
Fix 7818

PatriceJiang:
add typecheck before invoke callback


PatriceJiang:
refer https://github.com/cocos-creator/fireball/issues/7818

PR #1371, opened 2018-07-05
[feature] localStorage add api getKey and length

W-MTZing:


drelaptop:
look good, please merge @minggo

PR #1397, opened 2018-07-18
fix build crash on android when enable AnySDK

drelaptop:
for issues

https://github.com/cocos-creator/fireball/issues/7921

1. correct init order for AnySDK JNI environment
2. restore AnySDK on template

### DIFF
--- a/cocos/platform/android/jni/JniImp.cpp
+++ b/cocos/platform/android/jni/JniImp.cpp
@@ -104,6 +104,8 @@ namespace
     }
 }
 
+void cocos_jni_env_init (JNIEnv* env);
+
 Application* cocos_android_app_init(JNIEnv* env, int width, int height);
 
 extern "C"
@@ -111,6 +113,7 @@ extern "C"
     JNIEXPORT jint JNI_OnLoad(JavaVM *vm, void *reserved)
     {
         JniHelper::setJavaVM(vm);
+        cocos_jni_env_init(JniHelper::getEnv());
         return JNI_VERSION_1_4;
     }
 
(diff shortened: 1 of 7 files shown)


cocos/platform/android/jni/JniImp.cpp
```diff
@@ -104,13 +104,16 @@ namespace
...
+void cocos_jni_env_init (JNIEnv* env);
+
 Application* cocos_android_app_init(JNIEnv* env, int width, int height);
 
 extern "C"
 {
     JNIEXPORT jint JNI_OnLoad(JavaVM *vm, void *reserved)
     {
         JniHelper::setJavaVM(vm);
+        cocos_jni_env_init(JniHelper::getEnv());
```
drelaptop: the `JNI_OnLoad` is called when load native library, and `cocos_android_app_init `was called at here when creator 1.9.x, AnySDK init JNI environment at this moment.

but on creator2.0, `cocos_android_app_init` was changed to another location, I think the reason is to get screen width and height

templates/js-template-default/frameworks/runtime-src/proj.android-studio/app/jni/hellojavascript/main.cpp
```diff
@@ -41,23 +41,29 @@ using namespace cocos2d;
...
 #if PACKAGE_AS
-    JavaVM* vm;
+    JavaVM *vm;
     env->GetJavaVM(&vm);
     PluginJniHelper::setJavaVM(vm);
 #endif
+}
+
+//called when onSurfaceCreated()
+Application* cocos_android_app_init (JNIEnv* env, int width, int height) {
```
drelaptop: add `cocos_jni_env_init` ,and call it at `JNI_OnLoad()`, so that we can init the AnySDK plugin earlier, like creator 1.9.x did.

now `cocos_android_app_init ` was called at `onSurfaceCreated()`. it was called two late, order like:
```
->onCreate(Acticity)
->onStart(Acticity)
->onResume(Acticity)
->setScreenWidthAndHeight(GLSurfaceView.Renderer ,Get Screen Width and Height here)
->onSurfaceCreated(GLSurfaceView.Renderer, native init here, include screen width and height info)
```

PR #1440, opened 2018-08-07
fix two issues:

minggo:
if return type is string, and return null from java to js, then it will crash
if the parameter is string, and pass null from js to java, then java will get "null" instead of null


minggo:
@wuzhiming 等这个 PR 合并后你可以试一下
@pandamicro

PR #1454, opened 2018-08-13
win32: fix invoke update() multi times with same dt

PatriceJiang:
ref http://forum.cocos.com/t/creator2-0-0-bug/64303

`update()` can be invoked multiple times with same `dt`, it leads to `Timer` runs about 5 times faster than the real clock.


PatriceJiang:
pls, review if available @minggo

PR #1528, opened 2018-09-29
add video player source for ios and android.

xianyinchen:


pandamicro:
https://github.com/cocos-creator/2d-tasks/issues/639

drelaptop:
基于此，重提 PR #1531

PR #1589, opened 2018-11-08
update gradle to 4.4, plugin 3.1.4

drelaptop:
- 原 gradle 插件是 android studio 3.0 的，和 ndk-r18 搭配在一起有已知问题:
https://android.googlesource.com/platform/ndk/+/ndk-release-r18/CHANGELOG.md

- ndk18 与 gradle plugin 3.0 的 issue 是：
> This version of the NDK is incompatible with the Android Gradle plugin version 3.0 or older. If you see an error like No toolchains found in the NDK toolchains folder for ABI with prefix: mips64el-linux-android, update your project file to use plugin version 3.1 or newer. You will also need to upgrade to Android Studio 3.1 or newer.

- 官方建议更新至 3.1+ ，现更新为 3.1.4 对应（Android Studio 3.1.4），如果使用 Android Studio 3.2 不要根据提示更新 Gradle 至 android plugin 3.2（gradle 4.6），因为 gradle 4.6 有已知 issues 会导致资源打包结果为空。

- gradle 4.6 已知 issues 是：

https://developer.android.com/studio/known-issues
> Modifying variant outputs at build time may not work
Using the Variant API to manipulate variant outputs is broken with the new plugin. It still works for simple tasks, such as changing the APK name during build time, as shown below:

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/CanvasRenderingContext2DImpl.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/CanvasRenderingContext2DImpl.java
@@ -191,7 +191,7 @@ public class CanvasRenderingContext2DImpl {
         if(obliqueFont) {
             paint.setTextSkewX(_sApproximatingOblique);
         }
-        if(smallCapsFontVariant && Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+        if(smallCapsFontVariant && Build.VERSION.SDK_INT >= 21) {
             paint.setFontFeatureSettings("smcp");
         }
         return paint;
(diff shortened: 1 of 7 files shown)


cocos/platform/android/java/src/org/cocos2dx/lib/CanvasRenderingContext2DImpl.java
```diff
@@ -191,7 +191,7 @@ private static TextPaint newPaint(String fontName, int fontSize, boolean enableB
         if(obliqueFont) {
             paint.setTextSkewX(_sApproximatingOblique);
         }
-        if(smallCapsFontVariant && Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+        if(smallCapsFontVariant && Build.VERSION.SDK_INT >= 21) {
```
drelaptop: 用户可能会用 SDK-19 编译，没有内置变量

> 需要区别使用 SDK-27，指定 API-Level 16

drelaptop: fix https://github.com/cocos-creator/cocos2d-x-lite/issues/1588

PR #1593, opened 2018-11-13
fix low API Level java compile error on Android

drelaptop:
fix https://github.com/cocos-creator/2d-tasks/issues/714


jareguo:
修复安卓 API Level 低于 21 时可能编译失败的问题

PR #1632, opened 2018-12-18
fix dragonbones shake bug

sunnylanwanjun:
https://forum.cocos.com/t/cocos-creator-v2-0-7/70876/37?u=jare
https://forum.cocos.com/t/v2-0-7-alpha/70898

### DIFF
--- a/cocos/editor-support/dragonbones/armature/Slot.cpp
+++ b/cocos/editor-support/dragonbones/armature/Slot.cpp
@@ -239,10 +239,10 @@ void Slot::_updateDisplayData()
                 _pivotY += _helpPoint.y;
             }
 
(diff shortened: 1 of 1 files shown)


cocos/editor-support/dragonbones/armature/Slot.cpp
```diff
@@ -239,10 +239,10 @@ void Slot::_updateDisplayData()
                 _pivotY += _helpPoint.y;
             }
 
-            if (!DragonBones::yDown)
-            {
+            //if (!DragonBones::yDown)
```
pandamicro: 我看 yDown 是 public static 啊，直接修改不就好了

PR #1641, opened 2018-12-24
improve get system language code for windows

knoxHuang:
完善 windows 平台上获取完整的系统字体


drelaptop:
关联 https://github.com/cocos-creator/cocos2d-x-lite/pull/1638

PR #1642, opened 2018-12-25
add Graph API bindings for Facebook PC SDK

drelaptop:


drelaptop:
please review @wuzhiming

PR #1679, opened 2019-02-15
fixed the compressed texture error

ColinCollins:
Re: https://github.com/cocos-creator/example-cases/pull/671
copy image data size doesn't match the compress data size
修复 2.1 compress texture 渲染报错问题。
因为之前提交代码不够严谨，导致 imgInfo->length 每次都会重新计算成 宽高 * signed_bytes


drelaptop:
@W-MTZing 留意一下这个的改动，考虑 runtime 是否需要

PR #1723, opened 2019-04-18
`WsThreadHelper::__defaultProtocols` are not zeroed properly

id-ilych:


jareguo:
Also sync with https://github.com/cocos2d/cocos2d-x/pull/18643

PR #1757, opened 2019-05-31
fix spine will crash when load/unload spine res frequently

random634:
合并请求到v2.1.2


sunnylanwanjun:
已经帮你cherry pick 到这个pr了，谢谢提交
https://github.com/cocos-creator/cocos2d-x-lite/pull/1758

jareguo:
嗯，那就不用重复提交啦

PR #1761, opened 2019-06-04
fix memory leak in loading font on iOS/mac

minggo:


minggo:
sync https://github.com/cocos-creator/cocos2d-x-lite/pull/1752

PR #1764, opened 2019-07-01
improve convert draw y point calculate for 2d-tasks/issues/1378 (#1739)

jareguo:
* improve convert draw y point calculate for 2d-tasks/issues/1378

* update comment

* improve convert draw y point calculate for android

* improve convert draw y point calculate for win32


jareguo:
之前打算让用户统一升级到 2.2，不过 2.2 拖太久了，改动太大，来不及让 2.1 用户升级了

PR #1788, opened 2019-07-31
 [Android] fix an audio bug

unace:
[Android] fix an audio bug that audios are not paused when locking screen while notification drawer is showing. Technically they are paused however they are resumed right after paused even though the screen locked.

repro steps:

1. play audio
2. swipe down the notification drawer
3. lock the screen

### DIFF
--- a/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
+++ b/cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
@@ -73,6 +73,8 @@ public abstract class Cocos2dxActivity extends Activity implements Cocos2dxHelpe
     private Cocos2dxWebViewHelper mWebViewHelper = null;
     private boolean hasFocus = false;
     private Cocos2dxEditBox mEditBox = null;
+    private boolean gainAudioFocus = false;
+    private boolean paused = true;
 
     // DEBUG VIEW BEGIN
     private LinearLayout mLinearLayoutForDebugView;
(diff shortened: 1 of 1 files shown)


minggo:
It is strange that https://github.com/cocos-creator/cocos2d-x-lite/pull/1785 is merged, why this PR also has these modifications.

minggo:
Ok, i got it, you use the same branch without synchronizing the codes.

cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxActivity.java
```diff
@@ -355,7 +371,7 @@ public void onWindowFocusChanged(boolean hasFocus) {
     }
 
     private void resumeIfHasFocus() {
-        if(hasFocus) {
+        if(hasFocus && !paused) {
```
minggo: Why it is needed? `resumeIfHasFocus()` is invoked by `onResome()`, and `paused` is set to false.

unace: in the scenario that I described, below steps happen.
1. onWindowFocusChanged(false) is called when swiping down notification drawer.
2. onPause() is called when locking the screen. -> This pauses all audios.
3. onWindowFocusChanged(true) is called right after 2. -> call resumeIfHasFocus() -> This resumes all audios even though the screen has been locked.

So it seems to be needed to check pause status.

I verified this scenario on Google Pixel XL and Google Pixel 2.

detail Cocos2dxActivity logs
D/Cocos2dxActivity: onWindowFocusChanged() hasFocus=false
D/Cocos2dxActivity: onPause()
D/Cocos2dxActivity: onWindowFocusChanged() hasFocus=true
D/Cocos2dxActivity: onWindowFocusChanged() hasFocus=false

PR #1878, opened 2019-10-14
fix crash if invoked from non c++ context

minggo:


holycanvas:
关联issue：https://github.com/cocos-creator/2d-tasks/issues/1909

PR #1887, opened 2019-10-17
Fix skew issue

sunnylanwanjun:
engine：https://github.com/cocos-creator/engine/pull/5560/files
fireball：https://github.com/cocos-creator/fireball/pull/9162
issue: https://github.com/cocos-creator/2d-tasks/issues/1944

### DIFF
--- a/cocos/renderer/scene/NodeMemPool.hpp
+++ b/cocos/renderer/scene/NodeMemPool.hpp
@@ -42,6 +42,8 @@ struct TRS {
     float sx;
     float sy;
     float sz;
+    float skx;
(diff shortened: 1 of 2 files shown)


cocos/renderer/scene/NodeMemPool.hpp
```diff
@@ -42,6 +42,8 @@ struct TRS {
     float sx;
     float sy;
     float sz;
+    float skx;
+    float sky;
```
jareguo: 这样 TRS 开销就大了，能不能做成无损的？

PR #1937, opened 2019-11-18
fix videoplayer crash issue

minggo:


holycanvas:
https://forum.cocos.org/t/cocos-creator-v2-2-1/85555/333
https://forum.cocos.org/t/cocos-creator-v2-2-1/85555/342?u=endevil

PR #1947, opened 2019-11-26
send pause/resume event instead of background/foreground event

minggo:


holycanvas:
https://github.com/cocos-creator/2d-tasks/issues/2085

PR #1977, opened 2019-12-09
Fix the problem that the shadow is not displayed on native.

caryliu1999:
https://github.com/cocos-creator/engine/issues/5812

修复原生平台阴影没有正常显示。


caryliu1999:
https://github.com/cocos-creator/cocos2d-x-lite/pull/1966 youyou 这个PR修复了，此PR关闭

PR #1991, opened 2019-12-13
fix ios black screen after adding depth/stencil renderbuffer

Mee-gu:


cocos/renderer/gfx-gles2/GLES2Context.mm
```diff
@@ -80,22 +86,55 @@
...
+                break;
+            }
+            case GL_FRAMEBUFFER_UNSUPPORTED:
+            {
+                CC_LOG_ERROR("glCheckFramebufferStatus() - FRAMEBUFFER_UNSUPPORTED");
+                break;
+            }
+            default:;
+        }
+        return false;
```
minggo: 返回之前得删除已经分配得资源

Mee-gu: got it.

PR #1999, opened 2019-12-20
Fix Blob request and support mac compilation

pandamicro:


pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/1838

PR #2013, opened 2019-12-24
windows port

YunHsiao:


pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/1838

PR #2060, opened 2020-01-09
:bug: fix crash when launch from icon after launch from installer

SnowflakeTQ:
Hi

I just found a crash bug, it can be reproduced with following steps.

1. install app with android installer, DO NOT click "DONE"
2. click the "Open" button to launch the app
3. make the app enter background
4. click the icon to resume the app

when you click the icon to resume the app, it will recreate a new activity, `isTaskRoot` is `false`, so the new activity will be finish directly.
```java
        if (!isTaskRoot()) {
            // Android launched another instance of the root activity into an existing task
            //  so just quietly finish and go away, dropping the user back into the activity
            //  at the top of the stack (ie: the last state of this task)
            finish();
            Log.w(TAG, "[Workaround] Ignore the activity started from icon!");
            return;
        }
```
the `onDestroy` will be called, and will be crash with this line
```java
Cocos2dxHelper.unregisterBatteryLevelReceiver(this);
```

by the way, the following code should be called after `isTaskRoot` checker.
```
Utils.setActivity(this);
```


SnowflakeTQ:
any updates?

minggo:
Thanks, it looks good to me.

holycanvas:
Could you make another PR to Branch:v2.3.0-release

PR #2106, opened 2020-01-21
Fix crash when playing audio under mac/ios

sanikoyes:
use `_threadAudioPlayers` instead of `_audioPlayers` for audio thread callback `AudioEngineImpl::myAlSourceNotificationCallback` iterate all playing audio players.
Avoid crashes when accessing `_audioPlayers` in the main and child threads when playing multiple audio.


PatriceJiang:
It is weird to maintain a redundant `list` here. I suggest that the crash can be avoided by acquiring lock during accessing `_audioPlayers`. @sanikoyes

jareguo:
@sanikoyes Hi are you there?

holycanvas:
Thanks anyway

PR #2212, opened 2020-02-28
fix compile issue due to code refine

Mee-gu:


cocos/base/ccUtils.cpp
```diff
@@ -28,6 +28,10 @@ THE SOFTWARE.
 #include "base/base64.h"
 #include "platform/CCFileUtils.h"
 
+#if CC_TARGET_PLATFORM == CC_PLATFORM_IOS || CC_TARGET_PLATFORM == CC_PLATFORM_MAC
```
minggo: 宏不对吧

PR #2240, opened 2020-03-06
update submodule, dont generate api docs 

PatriceJiang:
depends on https://github.com/cocos-creator/bindings-generator/pull/52


minggo:
冲突了

PR #2274, opened 2020-03-13
convert to RGBA8

minggo:


minggo:
https://github.com/cocos-creator/3d-tasks/issues/2267

PR #2296, opened 2020-03-18
Avoid accessing se::Object in non-main thread

PatriceJiang:
ref https://github.com/cocos-creator/2d-tasks/issues/2640

avoid copying in non-main thread


PatriceJiang:
@minggo 麻烦review下

minggo:
3d 和 3d-gfx 也有类似的问题，得同步一下。

PR #2327, opened 2020-03-24
use gles3 for all platforms

minggo:
engine: https://github.com/cocos-creator/engine/pull/6406
issue: https://github.com/cocos-creator/3d-tasks/issues/2692

@PatriceJiang windows 可以去除 GLES2 相关的代码编译以减少编译时间。


PatriceJiang:
OK

PR #2354, opened 2020-04-07
improve metal implementation

minggo:


minggo:
https://github.com/cocos-creator/3d-tasks/issues/2449
engine: https://github.com/cocos-creator/engine/pull/6500
jsb-adapter: https://github.com/cocos-creator-packages/jsb-adapter/pull/276

PR #2363, opened 2020-04-09
fix crash that buffer is not bound

minggo:


minggo:
https://github.com/cocos-creator/cocos2d-x-lite/pull/2363

PR #2385, opened 2020-04-20
V2.4.0 add win size

wuzhiming:
add win size setting support for windows and mac platform
https://github.com/cocos-creator/2d-tasks/issues/149


wuzhiming:
@minggo @PatriceJiang please review

templates/js-template-default/frameworks/runtime-src/Classes/NativeConfig.h
```diff
@@ -0,0 +1,31 @@
+/****************************************************************************
+ Copyright (c) 2018 Xiamen Yaji Software Co., Ltd.
```
minggo, line 2: 2018  -> 2020

PR #2396, opened 2020-04-22
download-deps.py auto/manual select repo

PatriceJiang:


PatriceJiang:
@wuzhiming pls review if you are available

PR #2407, opened 2020-04-27
[bugfix] crash during global object destructuring / 3d

PatriceJiang:
ref https://github.com/cocos-creator/cocos2d-x-lite/pull/2405


minggo:
需要同步到 3d-gfx 分支

PatriceJiang:
已同步

PR #2459, opened 2020-05-18
3dgfx JNI improve thread safe

PatriceJiang:
task: https://github.com/cocos-creator/2d-tasks/issues/2819 

ref: https://github.com/cocos2d/cocos2d-x/pull/18189


minggo:
Now `jni env` is not attach to current thread, it is cached, so i think this PR can not make it thread safe thoroughly.

minggo:
I don't know how this PR resolve thread safe. It seems there is not difference compared to current implementation.

PatriceJiang:
The issue is described here https://github.com/cocos2d/cocos2d-x/issues/16581

minggo:
Oh, thanks. I forgot to fix it.

PR #2461, opened 2020-05-18
gfx-vk: all c++ test cases cleared

YunHsiao:
https://github.com/cocos-creator/3d-tasks/issues/2450


cocos/renderer/core/gfx/GFXDevice.h
```diff
@@ -67,8 +67,8 @@ class CC_CORE_API GFXDevice : public Object {
   CC_INLINE bool hasFeature(GFXFeature feature) const { return _features[static_cast<uint8_t>(feature)]; }
   CC_INLINE void defineMacro(const String& macro, const String& value) { _macros[macro] = value; }
   CC_INLINE void setReverseCW(bool reverseCW) { _reverseCW = reverseCW; }
-    CC_INLINE float getMinClipZ() const { return _minClipZ; }
-    CC_INLINE int getProjectionSignY() const { return _projectionSignY; }
+  CC_INLINE float getMinClipZ() const { return _minClipZ; }
+  CC_INLINE float getProjectionSignY() const { return _projectionSignY; }
```
minggo, line 71: Is there any reason to use float?

YunHsiao: these factors are most commonly seen as parameters to projection matrices or vertex datas, both of which are of floating point type

cocos/renderer/core/gfx/GFXCommandBuffer.h
```diff
@@ -13,7 +13,7 @@ class CC_CORE_API GFXCommandBuffer : public GFXObject {
 public:
   virtual bool initialize(const GFXCommandBufferInfo& info) = 0;
   virtual void destroy() = 0;
-  virtual void begin() = 0;
+  virtual void begin(GFXRenderPass* renderPass = nullptr, uint subpass = 0, GFXFramebuffer* frameBuffer = nullptr) = 0;
```
minggo, line 16: What happened if only some parameters are passed, for example,
```c++
begin(renderPass);
```

YunHsiao: Then it would mean this secondary command buffer will be executed in the subpass 0 of a renderpass compatible with `renderPass`, with any proper framebuffer.

cocos/renderer/gfx-vulkan/VKDevice.cc
```diff
@@ -247,19 +296,22 @@ void CCVKDevice::destroy()
...
     CC_SAFE_DESTROY(_cmdAllocator);
     CC_SAFE_DESTROY(_queue);
+    CC_SAFE_DESTROY(_stagingBuffer);
     CC_SAFE_DESTROY(_window);
     CC_SAFE_DESTROY(_renderPass);
     CC_SAFE_DELETE(_gpuSemaphorePool);
     CC_SAFE_DELETE(_gpuFencePool);
 
-    for (auto textureView : _depthStencilTextureViews)
+    for (CCVKTextureView* &textureView : _depthStencilTextureViews)
```
minggo, line 305: The usage is weird. Can not just use `CCVKTextureView*`?

cocos/renderer/gfx-vulkan/VKDevice.h
```diff
@@ -69,6 +68,9 @@ class CC_VULKAN_API CCVKDevice : public GFXDevice
     std::vector<CCVKTextureView*> _depthStencilTextureViews;
     std::vector<CCVKTexture*> _depthStencilTextures;
     CCVKRenderPass* _renderPass;
+    CCVKBuffer* _stagingBuffer;
```
minggo, line 71: It is better to assign default value here.

external/config.json
```diff
@@ -1,5 +1,5 @@
 {
-    "version": "v3d-gfx-12",
+    "version": "v3d-gfx-14",
```
minggo, line 2: Sorry, `v3d-gfx-14` is deleted. Please use `v3d-gfx-13` instead.

YunHsiao:
> The usage is weird. Can not just use CCVKTextureView*?

It's there so that `CC_SAFE_DESTROY` could properly reset the pointer back to null, but I guess that doesn't matter much since we are clearing right after that.

cocos/renderer/core/gfx/GFXDevice.h
```diff
@@ -105,7 +105,7 @@ class CC_CORE_API GFXDevice : public Object {
     uint _shaderIdGen = 0;
     std::unordered_map<String, String> _macros;
     float _minClipZ = -1;
-    int _projectionSignY = 1;
+    float _projectionSignY = 1;
```
minggo, line 108: Use 1.0f is better as the type is changed.

cocos/renderer/core/gfx/GFXCommandBuffer.h
minggo: Got it, thanks. But what happened if using like this:
```c++
begin(renderPass, 1);
```

As frame buffer is not passed, but the sub pass is 1. Will the application crash?

Mee-gu, line 16: need modify MTLCommandBuffer::begin as well.

minggo:
> > The usage is weird. Can not just use CCVKTextureView*?
> 
> It's there so that `CC_SAFE_DESTROY` could properly reset the pointer back to null, but I guess that doesn't matter much since we are clearing the vector right after that.

CC_SAFE_DESTROY is a macro, it is not a function.

cocos/renderer/core/gfx/GFXCommandBuffer.h
Mee-gu: it seems that all changes does not apply to metal

cocos/renderer/gfx-gles2/GLES2Commands.cc
```diff
@@ -1989,17 +1989,14 @@ void GLES2CmdFuncCopyBuffersToTexture(GLES2Device* device, uint8_t* const* buffe
...
-        if (!isCompressed) {
-          glTexSubImage2D(GL_TEXTURE_2D, m, region.texOffset.x, region.texOffset.y, w, h, gpuTexture->glFormat, gpuTexture->glType, (GLvoid*)buff);
-        } else {
-          GLsizei memSize = (GLsizei)GFXFormatSize(gpuTexture->format, w, h, 1);
-          glCompressedTexSubImage2D(GL_TEXTURE_2D, m, region.texOffset.x, region.texOffset.y, w, h, gpuTexture->glFormat, memSize, (GLvoid*)buff);
-        }
-
-        w = std::max(w >> 1, 1U);
-        h = std::max(h >> 1, 1U);
+      uint8_t* buff = buffers[n++];
```
Mee-gu, line 1992: need sync to metal

cocos/renderer/core/CoreDef.h
```diff
@@ -172,6 +172,35 @@
...
+// CLANG ENABLE/DISABLE WARNING DEFINITION
+#	define CC_DISABLE_WARNINGS()                                \
+        _Pragma("clang diagnostic push")                        \
+            _Pragma("clang diagnostic ignored \"-Wall\"")       \
+                _Pragma("clang diagnostic ignored \"-Wextra\"") \
+                    _Pragma("clang diagnostic ignored \"-Wtautological-compare\"")
+
+#	define CC_ENABLE_WARNINGS() \
+		_Pragma("clang diagnostic pop")
+#endif
```
minggo, line 202: Cool, i like these macros.

cocos/renderer/gfx-vulkan/VKCommandBuffer.cc
```diff
@@ -51,7 +54,7 @@ void CCVKCommandBuffer::destroy()
     _status = GFXStatus::UNREADY;
 }
 
-void CCVKCommandBuffer::begin()
+void CCVKCommandBuffer::begin(GFXRenderPass* renderPass, uint subpass, GFXFramebuffer* frameBuffer)
```
Mee-gu, line 57: need sync to metal

cocos/renderer/gfx-vulkan/VKSPIRV.h
```diff
@@ -164,16 +164,16 @@ namespace
         /* .generalConstantMatrixVectorIndexing = */ 1,
     } };
 
-    const std::vector<unsigned int> GLSL2SPIRV(GFXShaderType type, String source, int vulkanMinorVersion = 2)
+    const vector<unsigned int>::type GLSL2SPIRV(GFXShaderType type, String source, int vulkanMinorVersion = 2)
```
Mee-gu, line 167: source is better to be const String&

cocos/renderer/core/gfx/GFXCommandBuffer.h
YunHsiao: I don't have the environment to check it so... I figured it would be better if someone else fill in those gaps.

YunHsiao:
> > > The usage is weird. Can not just use CCVKTextureView*?
> > 
> > 
> > It's there so that `CC_SAFE_DESTROY` could properly reset the pointer back to null, but I guess that doesn't matter much since we are clearing the vector right after that.
> 
> CC_SAFE_DESTROY is a macro, it is not a function.



> CC_SAFE_DESTROY is a macro, it is not a function.

macro or not it needs a reference to assign it back to null in a for loop like this I think..?

cocos/renderer/core/gfx/GFXCommandBuffer.h
Mee-gu: please let me know if you need me to fix the metal fix, or otherwise CI will failed

minggo:
> > > > The usage is weird. Can not just use CCVKTextureView*?
> > > 
> > > 
> > > It's there so that `CC_SAFE_DESTROY` could properly reset the pointer back to null, but I guess that doesn't matter much since we are clearing the vector right after that.
> > 
> > 
> > CC_SAFE_DESTROY is a macro, it is not a function.
> 
> macro or not it needs a reference to assign it back to null in a for loop like this I think..?

macro doesn't need reference, it just assign to the variable. For example
```c++
auto gfxBuffer = device->createBuffer(...);
CC_SAFE_DESTROY(gfxBuffer);
```

after pre-process, the codes will change to
```
auto gfxBuffer = device->createBuffer(...);
if (gfxBuffer) {
    gfxBuffer->destroy();
    gfxBuffer = nullptr;
}
```
So don't need to use reference here.

cocos/renderer/core/gfx/GFXCommandBuffer.h
YunHsiao: No, the framebuffer is optional as per vulkan API

minggo: Good to know, thanks.

YunHsiao: I've pushed all the changes I can think of, but I'm working in the dark here, so please verify the changes and if needed, submit a fix PR to this branch.

YunHsiao:
> > > > > The usage is weird. Can not just use CCVKTextureView*?
> > > > 
> > > > 
> > > > It's there so that `CC_SAFE_DESTROY` could properly reset the pointer back to null, but I guess that doesn't matter much since we are clearing the vector right after that.
> > > 
> > > 
> > > CC_SAFE_DESTROY is a macro, it is not a function.
> > 
> > 
> > macro or not it needs a reference to assign it back to null in a for loop like this I think..?
> 
> macro doesn't need reference, it just assign to the variable. For example
> 
> ```c++
> auto gfxBuffer = device->createBuffer(...);
> CC_SAFE_DESTROY(gfxBuffer);
> ```
> 
> after pre-process, the codes will change to
> 
> ```c++
> auto gfxBuffer = device->createBuffer(...);
> if (gfxBuffer) {
>     gfxBuffer->destroy();
>     gfxBuffer = nullptr;
> }
> ```
> 
> So don't need to use reference here.

Right but you see the situation here is a little different:
```
    for (CCVKTextureView* textureView : _depthStencilTextureViews)
    {
        CC_SAFE_DESTROY(textureView);
    }
```
which will expands to:
```
    for (CCVKTextureView* textureView : _depthStencilTextureViews)
    {
        if (textureView) { textureView->destroy(); textureView = null; }
    }
```
wtih the for loop indirection, I don't think the textureView handle in the vector can be reset correctly . But anyways that's not an issue since we are clearing the whole vector right after that.

minggo:
> > > > > > The usage is weird. Can not just use CCVKTextureView*?
> > > > > 
> > > > > 
> > > > > It's there so that `CC_SAFE_DESTROY` could properly reset the pointer back to null, but I guess that doesn't matter much since we are clearing the vector right after that.
> > > > 
> > > > 
> > > > CC_SAFE_DESTROY is a macro, it is not a function.
> > > 
> > > 
> > > macro or not it needs a reference to assign it back to null in a for loop like this I think..?
> > 
> > 
> > macro doesn't need reference, it just assign to the variable. For example
> > ```c++
> > auto gfxBuffer = device->createBuffer(...);
> > CC_SAFE_DESTROY(gfxBuffer);
> > ```
> > 
> > 
> > after pre-process, the codes will change to
> > ```c++
> > auto gfxBuffer = device->createBuffer(...);
> > if (gfxBuffer) {
> >     gfxBuffer->destroy();
> >     gfxBuffer = nullptr;
> > }
> > ```
> > 
> > 
> > So don't need to use reference here.
> 
> Right but you see the situation here is a little different:
> 
> ```
>     for (CCVKTextureView* textureView : _depthStencilTextureViews)
>     {
>         CC_SAFE_DESTROY(textureView);
>     }
> ```
> 
> which will be expanded to:
> 
> ```
>     for (CCVKTextureView* textureView : _depthStencilTextureViews)
>     {
>         if (textureView) { textureView->destroy(); textureView = null; }
>     }
> ```
> 
> wtih the for loop indirection, I don't think the textureView handle in the vector can be reset correctly(hence the reference like before). But anyways that's not an issue since we are clearing the whole vector right after that, so I removed the reference.

Yep, you are right. May be we need two macros:
- CC_SAFE_DESTROY, will not set the variable to null
- CC_SAFE_DESTROY_NULL, will set the pointer to null as it is now

CMakeLists.txt
```diff
@@ -655,6 +656,69 @@ if(CC_USE_METAL)
     )
 endif()
 
+if(CC_USE_VULKAN)
+
+    add_definitions("-DVK_NO_PROTOTYPES")
```
PatriceJiang, line 661: `add_definitions` 是全局的, 比较旧的写法. 建议使用 `target_compile_definitions(cocos2d PUBLIC )`, 在  `add_library(cocos2d` 之后)

PR #2464, opened 2020-05-19
Add .browserslist.rc

shrinktofit:
This PR adds a new file `.browserslistrc` which provides information about the [browserslist](https://github.com/browserslist/browserslist) query.

@pandamicro  @minggo  This should be merged into version 1.2, but as this PR creating the v1.2 branch has not been created yet. Please redirect the merge target to that version branch.

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,12 @@
+# This file provides [browserslist](https://github.com/browserslist/browserslist) query.
+# Syntax and meaning of this file obeys browserslist's document.
+#
+# Lite currently use V8 as its backend.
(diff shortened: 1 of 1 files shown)


minggo:
I think merging into 3d is ok though it is not used.

shrinktofit:
> I think merging into 3d is ok though it is not used.

Starting from 1.2, Editor would test if `.browserslist.rc` were existed in Lite's repository. So we might not put it into current 3d branch(which is used by 1.1 and before).

shrinktofit:
As @minggo and @pandamicro  confirmed `3d` is the working branch of 1.2, I have now marked this PR as  ready for review.

PR #2467, opened 2020-05-19
catch exception when `ScriptEngine::evalString` / 3d

PatriceJiang:
- Catch exception when `evalString`
- Fix cmake resource path / win32


minggo:
Send to v1.1.1 as 3d is used for v1.2?

PatriceJiang:
Re-commit in https://github.com/cocos-creator/cocos2d-x-lite/pull/2493

PR #2470, opened 2020-05-19
Use v8 exposed GC function / 3d

PatriceJiang:
Set v8 flags `--expose-gc-as`

Invoke v8 builtin garbage collection API directly, no tricks. 

https://github.com/v8/v8/blob/4f9b30b0145fcdda34e7012c7633cc95f6d3bbdf/src/flags/flag-definitions.h#L1114


minggo:
Send to v1.1.1 as 3d branch is used for v1.2?

PR #2478, opened 2020-05-21
`Pass::generateDefinesKey()` use ordered map to generate hash

PatriceJiang:


cocos/renderer/renderer/Pass.h
```diff
@@ -175,7 +176,7 @@ class Pass : public Ref
     Pass* _parent = nullptr;
     
     std::unordered_map<size_t, Technique::Parameter> _properties;
-    ValueMap _defines;
```
PatriceJiang, line 178: The iteration order of `std::unordered_map` is not stable. `Pass::generateDefinesKey()` can generate different hash for the same `defines` settings.

cocos/2d/CCLabelLayout.cpp
```diff
@@ -317,6 +317,10 @@ namespace cocos2d {
                     effect->copy(templateEffect);
                     assembler->updateEffect(groupIndex, effect);
                 }
+                
+                if(effect->getPasses().at(0)->getDefinesHash() != templateEffect->getPasses().at(0)->getDefinesHash()){
```
PatriceJiang, line 321: Update effect when defines changes.

PR #2479, opened 2020-05-21
more vulkan adaptations

YunHsiao:


cocos/renderer/core/gfx/GFXFence.h
```diff
@@ -0,0 +1,26 @@
+#ifndef CC_CORE_GFX_FENCE_H_
+#define CC_CORE_GFX_FENCE_H_
```
minggo, line 2: Can just use `#pragma once`

cocos/renderer/core/gfx/GFXQueue.h
```diff
@@ -13,7 +13,7 @@ class CC_CORE_API GFXQueue : public GFXObject {
 public:
   virtual bool initialize(const GFXQueueInfo& info) = 0;
   virtual void destroy() = 0;
-  virtual void submit(const std::vector<GFXCommandBuffer*>& cmd_buffs) = 0;
+  virtual void submit(const std::vector<GFXCommandBuffer*>& cmd_buffs, GFXFence* fence = nullptr) = 0;
```
minggo, line 16: It is better not use default parameter as possible.

cocos/renderer/core/gfx/GFXFence.h
YunHsiao: I think [it's not C++ standard](https://en.wikipedia.org/wiki/Pragma_once), so I prefer it this way

cocos/renderer/core/gfx/GFXQueue.h
YunHsiao: would function overloads be better for this purpose?

minggo: Yep, i think it is better here.

cocos/renderer/core/gfx/GFXFence.h
minggo: cocos2d-x uses it like this for years. Almost all compilers support it, and as i know, unreal use it too. It is ok you like this.

cocos/renderer/gfx-gles2/GLES2Fence.h
```diff
@@ -0,0 +1,25 @@
...
+class CC_GLES2_API GLES2Fence : public GFXFence {
+public:
+    GLES2Fence(GFXDevice* device);
+    ~GLES2Fence();
+
+public:
+    bool initialize(const GFXFenceInfo& info);
+    void destroy();
+    void wait();
+    void reset();
```
minggo, line 17: It is better to add `virtual` and `override`.

cocos/renderer/gfx-gles2/GLES2Fence.h
```diff
@@ -0,0 +1,25 @@
...
+
+private:
+    GLES2GPUFence* _gpuFence;
```
minggo, line 20: It is better to add default value.

cocos/renderer/gfx-gles2/GLES2Fence.cc
```diff
@@ -0,0 +1,49 @@
...
+{
+}
+
+GLES2Fence::~GLES2Fence()
+{
+}
+
+bool GLES2Fence::initialize(const GFXFenceInfo &info)
+{
+    _gpuFence = CC_NEW(GLES2GPUFence);
```
minggo, line 19: It is better to check result.

cocos/renderer/gfx-gles2/GLES2Fence.cc
```diff
@@ -0,0 +1,49 @@
...
+NS_CC_BEGIN
+
+GLES2Fence::GLES2Fence(GFXDevice* device)
+    : GFXFence(device)
+{
+}
+
+GLES2Fence::~GLES2Fence()
+{
+}
```
minggo, line 15: It is better to invoke `destroy()`.

YunHsiao: I think we shouldn't do that here, the creator of GFX resources should be responsible for destroying it. Right now it's a matter of exit event (or lack of it there of), which should be passes to JS so that it can destroy them by the right time.

cocos/renderer/gfx-gles2/GLES2Fence.cc
YunHsiao: Other use cases didn't do that, so I kept the same convention here.

cocos/renderer/gfx-gles2/GLES2Fence.h
YunHsiao: All these functions are pure virual functions

minggo: Yep, add `virtual` to make maintaining more easier. Add `override` can let compiler to check the error. The existing implementation doesn't use it, i will modify them.

cocos/renderer/gfx-gles2/GLES2Fence.cc
minggo: Yep, we should modify them if found. Check memory allocation result is a good practice, it makes codes more robust.

dumganhar:
现在改为使用 SDL 了？

cocos/renderer/gfx-gles2/GLES2Fence.cc
minggo: Yep, indeed JS should make things work as expected. But invoke destroy() here doesn't make performance issue if it is already destroyed, and can make sure it is destroyed if it is not destroyed by JS to avoid memory leak.

I think the class or function itself make things more robust by itself as possible is a good practice.

YunHsiao: Generally yes, but then there's the destroy order issue, the gfx device & context should always be destroy last, relying on destructors to do this will immediately lead to many validation errors in vulkan backend.

minggo:
> 现在改为使用 SDL 了？

是的，用来处理 windows 系统相关。

PR #2487, opened 2020-05-26
remove unused code / 3dgfx

PatriceJiang:
- update `default` template
- remove `Android.mk` support for Android
- enable module configuration in cmake
- enable pre-compiled headers
- link resource for win32 projects
---
external update https://github.com/cocos-creator/cocos2d-x-lite-external/pull/98
editor https://github.com/cocos-creator/editor-3d/pull/3594

task: https://github.com/cocos-creator/3d-tasks/issues/2412


minggo:
So we can only keep default template and remove link template, right?

PatriceJiang:
yes

PR #2488, opened 2020-05-27
fix dynamic states

Mee-gu:


cocos/renderer/core/gfx/GFXCommandBuffer.h
```diff
@@ -32,7 +32,8 @@ class CC_CORE_API GFXCommandBuffer : public GFXObject {
   virtual void updateBuffer(GFXBuffer* buff, void* data, uint size, uint offset = 0) = 0;
   virtual void copyBufferToTexture(GFXBuffer* src, GFXTexture* dst, GFXTextureLayout layout, const GFXBufferTextureCopyList& regions) = 0;
   virtual void execute(const std::vector<GFXCommandBuffer*>& cmd_buffs, uint32_t count) = 0;
-  
+  virtual void enableDepthBias(bool enabled) { _depthBiasEnabled = enabled; }
```
minggo, line 35: I think use setDepthBias is better.

cocos/renderer/gfx-gles2/GLES2Commands.cc
```diff
@@ -1899,10 +1899,149 @@ void GLES2CmdFuncExecuteCmds(GLES2Device* device, GLES2CmdPackage* cmd_package)
...
+                        glPolygonOffset(cmd->depthBias.constant, cmd->depthBias.slope);
+                        cache->rs.depthBias = cmd->depthBias.constant;
+                        cache->rs.depthBiasSlop = cmd->depthBias.slope;
+                    }
+                    break;
+                case GFXDynamicState::BLEND_CONSTANTS:
+                    if ((cache->bs.blendColor.r != gpuPipelineState->bs.blendColor.r) ||
+                        (cache->bs.blendColor.g != gpuPipelineState->bs.blendColor.g) ||
+                        (cache->bs.blendColor.b != gpuPipelineState->bs.blendColor.b) ||
+                        (cache->bs.blendColor.a != gpuPipelineState->bs.blendColor.a))
```
minggo, line 1951: Can implement operator == and != in GFXColor to simplify the codes. And float comparison can not just use !=.

cocos/renderer/core/gfx/GFXCommandBuffer.h
Mee-gu: there is already a `setDepthBias ` to set depth bias

cocos/renderer/gfx-gles2/GLES2Commands.cc
Mee-gu: got it.

minggo:
I think we should support all dynamics that are supported on all platforms. As some dynamic states are on supported on metal, then can refer to molten implementation.

Mee-gu:
> I think we should support all dynamics that are supported on all platforms. As some dynamic states are on supported on metal, then can refer to molten implementation.

I will take a look.

PR #2501, opened 2020-05-28
add hash function for GFXInputAssembler and GFXFrameBuffer

minggo:
Corresponding JS engine PR is: https://github.com/cocos-creator/engine/pull/6732


minggo:
I am not sure if i need to add a hash util function.

PR #2512, opened 2020-06-01
fix some potential memleaks and null pointer dereferences

quink-black:


quink-black:
Ping for review, thanks!

cocos/scripting/js-bindings/manual/jsb_global.cpp
```diff
@@ -831,36 +831,42 @@ bool jsb_global_load_image(const std::string& path, const se::Value& callbackVal
     std::shared_ptr<se::Value> callbackPtr = std::make_shared<se::Value>(callbackVal);
 
     auto initImageFunc = [path, callbackPtr](const std::string& fullPath, unsigned char* imageData, int imageBytes, const std::string& errorMsg){
-        Image* img = new (std::nothrow) Image();
+        std::shared_ptr<uint8_t> imageDataGuard(imageData, free);
```
PatriceJiang, line 834: I suggest using `unique_ptr`

cocos/scripting/js-bindings/manual/jsb_global.cpp
```diff
@@ -831,36 +831,42 @@ bool jsb_global_load_image(const std::string& path, const se::Value& callbackVal
...
+                loadSucceed = img->initWithImageData(imageDataGuard.get(), imageBytes);
+                imageDataGuard = nullptr;
             }
             else
             {
                 loadSucceed = img->initWithImageFile(fullPath);
             }
 
-            struct ImageInfo* imgInfo = nullptr;
+            std::shared_ptr<ImageInfo> imgInfo;
```
PatriceJiang, line 863: use `unique_ptr`

cocos/scripting/js-bindings/manual/jsb_global.cpp
quink-black: It's not easy to capture unique_ptr with C++11, ref.

https://stackoverflow.com/questions/8236521/how-to-capture-a-unique-ptr-into-a-lambda-expression

We can do that after cocos update to C++14.

quink-black:
可以merge了吗？

PR #2514, opened 2020-06-02
update spine to 3.8.95+ / v2.4.0

PatriceJiang:
Update spine for native platforms
ref https://github.com/cocos-creator/engine/pull/6798

https://github.com/EsotericSoftware/spine-runtimes/issues/1685


PatriceJiang:
This will break compatibility with older versions of spine files.

PR #2524, opened 2020-06-03
travis add nodejs / 3dgfx 

PatriceJiang:


PatriceJiang:
@minggo please review if you are available

PR #2527, opened 2020-06-04
[bugfix] win32 drawtext prefix

PatriceJiang:
Win32 turns off processing of prefix characters while drawing text.

### DIFF
--- a/cocos/platform/win32/CCCanvasRenderingContext2D-win32.cpp
+++ b/cocos/platform/win32/CCCanvasRenderingContext2D-win32.cpp
@@ -402,7 +402,7 @@ private:
         {
             CC_BREAK_IF(text.empty());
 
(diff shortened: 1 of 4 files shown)


PatriceJiang:
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-drawtext 

### DT_NOPREFIX 

 Turns off processing of prefix characters. Normally, DrawText interprets the mnemonic-prefix character & as a directive to underscore the character that follows, and the mnemonic-prefix characters && as a directive to print a single &. By specifying`DT_NOPREFIX`, this processing is turned off. 
For example, 
Example:
-  input string: "A&bc&&d"
-  normal: "Abc&d" -
-  `DT_NOPREFIX`: "A&bc&&d"

PR #2532, opened 2020-06-05
Fix crash below android4.4 when the class cannot be found

deyongzhong:
If this class does not exist, devices below Android 4.4 may crash

### DIFF
--- a/cocos/platform/android/jni/JniHelper.cpp
+++ b/cocos/platform/android/jni/JniHelper.cpp
@@ -49,9 +49,10 @@ jclass _getClassID(const char *className) {
                                                    cocos2d::JniHelper::loadclassMethod_methodID,
                                                    _jstrClassName);
 
-    if (nullptr == _clazz) {
+    if (nullptr == _clazz || env->ExceptionCheck()) {
         LOGE("Classloader failed to find class of %s", className);
         env->ExceptionClear();
+        _clazz = nullptr;
     }
 
     env->DeleteLocalRef(_jstrClassName);


cocos/platform/android/jni/JniHelper.cpp
```diff
@@ -49,9 +49,10 @@ jclass _getClassID(const char *className) {
                                                    cocos2d::JniHelper::loadclassMethod_methodID,
                                                    _jstrClassName);
 
-    if (nullptr == _clazz) {
+    if (nullptr == _clazz || env->ExceptionCheck()) {
```
PatriceJiang, line 52: It seems that  the class is found and some exception occurs. Would you share what kind of exception did you met?

deyongzhong:
当这个类不存在的时候，在android 4.4以下设备,_clazz不为 nullptr,但是有异常,这个是bugly上收集到的异常报告。

java.lang.ClassNotFoundException: Didn't find class "com.me2zen.remoteconfig.RemoteConfigManager" on path: DexPathList[[

deyongzhong:
更新一下，5.1.1的设备也有这个问题。

PatriceJiang:
https://developer.android.com/training/articles/perf-jni
>  Many JNI calls can throw an exception, but often provide a simpler way of checking for failure. For example, if NewString returns a non-NULL value, you don't need to check for an exception. However, if you call a method (using a function like CallObjectMethod), you must always check for an exception, because the return value is not going to be valid if an exception was thrown.

holycanvas:
感谢

PR #2534, opened 2020-06-08
[bugfix] Update iOS compile flags for V8

PatriceJiang:
- Remove macro `V8_COMPRESS_POINTERS`
- Init `AlignedPointerInInternalField` for `gl` object.


PatriceJiang:
@minggo  麻烦合并下

PR #2599, opened 2020-06-18
vk: queues are optionally async

YunHsiao:


cocos/renderer/gfx-vulkan/VKCommands.cc
```diff
@@ -327,9 +327,7 @@ void CCVKCmdFuncUpdateBuffer(CCVKDevice *device, CCVKGPUBuffer *gpuBuffer, void
 }
 
 bool CCVKCmdFuncCreateTexture(CCVKDevice *device, CCVKGPUTexture *gpuTexture) {
-    if (!gpuTexture->width || !gpuTexture->height) {
-        return true;
-    }
+    if (!gpuTexture->width || !gpuTexture->height) return true;
```
Mee-gu, line 330: return false is better ?

YunHsiao: return false here will trigger the device.createTexture function to return nullptr, which is not what we want on this situation

Mee-gu: maybe when width or height is equals to 0, engine not trigger the device.createTexture since it not reasonable

YunHsiao: You mean just don't create them in the first place. That makes sense, but I think there's existing usage depending on this particular behavior. I'll see what I can do.

YunHsiao: I think the whole `VKCommands` util functions should handle whatever parameter is thrown at them gracefully, without any crashing or rejection, which is exactly what the GLES backend does. Based on that insight, whether to reject zero-dimension textures should be a delibrate choice, which we can easily toggle.

So for now, to keep the same behavior as other backends, I choose to allow zero-dimension textures to be created(but not actually allocating anything). This behavior can be easily changed by rejecting in the texture.initialize function, alongside with the format validation.

YunHsiao:
I think the whole VKCommands util functions should handle whatever parameter is thrown at them gracefully, without any crashing or rejection, which is exactly what the GLES backend does. Based on that design, whether to reject zero-dimension textures should be a delibrate choice, which we can easily toggle.

So for now, to keep the same behavior as other backends, I choose to allow zero-dimension textures to be created(but not actually allocating anything). This behavior can be easily changed by rejecting in the texture.initialize function, alongside with the format validation.

PR #2608, opened 2020-06-22
vk: fix indirect buffer

YunHsiao:


cocos/renderer/gfx-vulkan/VKGPUObjects.h
```diff
@@ -139,7 +139,11 @@ class CCVKGPUBuffer : public Object {
     uint stride = 0;
     uint count = 0;
     void *buffer = nullptr;
+
     bool isDrawIndirectByIndex = false;
+    vector<VkDrawIndirectCommand> indirectCmds;
```
Mee-gu, line 144: does it necessary to store these two cmds?

YunHsiao: It helps with debugging and save the work to allocate a vector every time we do a upload

PR #2657, opened 2020-07-25
vk: all buffer updates need staging buffer

YunHsiao:
or every buffer passed to update has to be kept around until command submission time, which is unacceptable.

### DIFF
--- a/cocos/base/CachedArray.h
+++ b/cocos/base/CachedArray.h
@@ -109,7 +109,7 @@ public:
                 return i;
             }
         }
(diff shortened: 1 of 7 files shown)


cocos/base/CachedArray.h
```diff
@@ -109,7 +109,7 @@ class CachedArray : public Object {
                 return i;
             }
         }
-        return -1;
+        return (uint)-1;
```
minggo, line 112: UINT_MAX?

PR #2669, opened 2020-08-03
fix editbox indicator jump to end on iOS

PPpro:
changeLog:
- 修复 iOS 端 editBox 编辑文字，光标自动跳转到末端的问题
- 修复 iOS 下 editBox 韩文联想被打断的问题 https://github.com/cocos-creator/2d-tasks/issues/2400


PPpro:
requested review from @minggo

PR #2720, opened 2020-08-25
Avoid upload a NULL image memory to the GL texture

xianyinchen:
Avoid upload a NULL image memory to the GL texture. Resolving some texture blur issue.


holycanvas:
最好查下原因，为啥会为空，上层应该处理了这种情况才是

xianyinchen:
Only happened on OPPO R7. 
Don't call setPremultiplyAlpha(true) when cc.macro.CLEANUP_IMAGE_CACHE is true.

PR #2795, opened 2020-09-16
get handle from JS

minggo:


minggo:
JS engine: https://github.com/cocos-creator/engine/pull/7389

PR #2799, opened 2020-09-17
do not define property of RenderView.visibility

minggo:
the property is defined in jsb adapter


minggo:
jsb adapter: https://github.com/cocos-creator/engine/pull/7398

PR #2960, opened 2020-10-29
fix staging buffer

Mee-gu:
ref issue: [3d-tasks/issues/4731](https://github.com/cocos-creator/3d-tasks/issues/4731#issue-730246095)


minggo:
有冲突了

PR #2999, opened 2020-11-09
fix GL invalid operation issue.

xianyinchen:
报错堆栈


xianyinchen:
问题一样，关闭
https://github.com/cocos-creator/cocos2d-x-lite/pull/2884

PR #3038, opened 2020-11-17
modify UNIFORM_SHADOWMAP's set to GLOBAL_SET

lawrence-liuyue:
modify UNIFORM_SHADOWMAP's set to GLOBAL_SET


Mee-gu:
more modification were made in [cocos2d-x-lite/pull/3045](https://github.com/cocos-creator/cocos2d-x-lite/pull/3045)

PR #3057, opened 2020-11-20
only manage triple buffer

minggo:


minggo:
在 https://github.com/cocos-creator/cocos2d-x-lite/pull/3056 修复了

PR #3058, opened 2020-11-20
support window event  on native desktop platform

PPpro:
re: https://github.com/cocos-creator/3d-tasks/issues/5152


PPpro:
@minggo please review

PatriceJiang (approved):

PR #3079, opened 2020-11-27
fix invalid blend target

Mee-gu:
fix [3d-tasks/issues/5362](https://github.com/cocos-creator/3d-tasks/issues/5362)


minggo:
提交到 preview 分支吧

Mee-gu:
push to preview branch

PR #3090, opened 2020-12-02
use float type color for particle to fix color lerp precision

caogtaa:
## Related Issue
https://github.com/cocos-creator/engine/issues/7206, this fixed JS version particle in early time.
This PR fix native version.

Particle color animation is not accurate for long living particle, use float type color to fix it.


holycanvas (approved):

holycanvas:
Thanks again

PR #3117, opened 2020-12-15
Modify the forward rendering process and remove UIFlow.

caryliu1999:


cocos/renderer/pipeline/Define.cpp
```diff
@@ -63,6 +47,38 @@ const gfx::UniformBlock UBOLocalBatched::LAYOUT = {
...
+const gfx::DescriptorSetLayoutBinding UBOCamera::DESCRIPTOR = {
+    UBOCamera::BINDING,
+    gfx::DescriptorType::UNIFORM_BUFFER,
+    1,
+    gfx::ShaderStageFlagBit::ALL,
+};
+const gfx::UniformBlock UBOCamera::LAYOUT = {
+    GLOBAL_SET,
+    UBOCamera::BINDING,
+    UBOCamera::NAME,
```
minggo, line 60: `UBOCamera::LAYOUT`, `UBOCamera::BINDING` and `UBOCamera::NAME` are all static member variables. `UBOCamera::BINDING` and `UBOCamera::NAME` may not be initialized when initializing `UBOCamera::LAYOUT`. So i think it is not a good way to use them here.

cocos/renderer/pipeline/Define.h
```diff
@@ -273,31 +274,40 @@ enum class CC_DLL RenderFlowTag {
...
+    static constexpr uint NATIVE_SIZE_OFFSET = UBOGlobal::SCREEN_SIZE_OFFSET + 4;
+    static constexpr uint COUNT = UBOGlobal::NATIVE_SIZE_OFFSET + 4;
     static constexpr uint SIZE = UBOGlobal::COUNT * 4;
     static constexpr uint BINDING = static_cast<uint>(PipelineGlobalBindings::UBO_GLOBAL);
     static const gfx::DescriptorSetLayoutBinding DESCRIPTOR;
     static const gfx::UniformBlock LAYOUT;
     static const String NAME;
 };
 
+struct CC_DLL UBOCamera : public Object {
```
minggo, line 286: I think UBOCamera does not have to inherit from Object, and struct inherit from Class is not a good.

cocos/renderer/pipeline/forward/ForwardStage.cpp
```diff
@@ -41,6 +41,7 @@ const RenderStageInfo &ForwardStage::getInitializeInfo() { return ForwardStage::
 ForwardStage::ForwardStage() : RenderStage() {
     _batchedQueue = CC_NEW(RenderBatchedQueue);
     _instancedQueue = CC_NEW(RenderInstancedQueue);
+    _uiPhase = CC_NEW(UIPhase);
```
troublemaker52025, line 44: UIPhase 应该做成基类Pipeline 里的默认实现呢？还是放在forward-pipeline 子类里实现？
如果forward-pipeline实现了，deffered-pipeline 是不是还要实现一遍？

minggo, line 44: where to delete it?

cocos/renderer/pipeline/forward/UIPhase.cpp
```diff
@@ -0,0 +1,52 @@
...
+            }
+        } else {
+            if (camera->visibility & batch->visFlags) {
+                visible = true;
+            }
+        }
+
+        if (!visible) continue;
+        const auto count = batch->passCount;
+        for (auto j = 0; j < count; j++) {
```
minggo, line 36: It is better not use auto for basic types.

cocos/bindings/auto/jsb_pipeline_auto.h
```diff
@@ -9,13 +9,10 @@
 #include "cocos/renderer/pipeline/forward/ForwardStage.h"
 #include "cocos/renderer/pipeline/shadow/ShadowFlow.h"
 #include "cocos/renderer/pipeline/shadow/ShadowStage.h"
-#include "cocos/renderer/pipeline/ui/UIFlow.h"
-#include "cocos/renderer/pipeline/ui/UIStage.h"
```
minggo, line 13: CI works now, do not have to change it manullay.

cocos/renderer/pipeline/forward/ForwardStage.cpp
caryliu1999: > where to delete it?

Fixed！

caryliu1999: > UIPhase 应该做成基类Pipeline 里的默认实现呢？还是放在forward-pipeline 子类里实现？
> 如果forward-pipeline实现了，deffered-pipeline 是不是还要实现一遍？

单独实现的，其他地方可以直接使用，直接遍历的 scene 中的 batch 数据

cocos/renderer/pipeline/Define.h
caryliu1999: > inherit

It seems to be unnecessary, should we change the other UBO structs together to test them?

cocos/renderer/pipeline/Define.cpp
caryliu1999: > `UBOCamera::LAYOUT`, `UBOCamera::BINDING` and `UBOCamera::NAME` are all static member variables. `UBOCamera::BINDING` and `UBOCamera::NAME` may not be initialized when initializing `UBOCamera::LAYOUT`. So i think it is not a good way to use them here.

UBOCamera::BINDING is initialized when constructed in the header file, and UBOCamera::NAME is initialized before UBOCamera::LAYOUT is assigned. Or it would have to be set up all at the time of construction in header file.

cocos/bindings/auto/jsb_pipeline_auto.h
caryliu1999: > CI works now, do not have to change it manullay.

Fixed!

cocos/renderer/pipeline/Define.h
minggo: Yep, i think we can create another issue to change them.

cocos/renderer/pipeline/Define.cpp
minggo: Yep, i think it can work now. But it depends on the sequence, if the sequence changes in future, then it will not work. And it may be hard to find the reason.

cocos/renderer/pipeline/forward/UIPhase.cpp
caryliu1999: > It is better not use auto for basic types.

fixed

PR #3119, opened 2020-12-16
fix some header include issue.

xianyinchen:


minggo:
What's the matter of the issue?

xianyinchen:
When I build it, I run into this issue.
image
image

minggo:
fixed in https://github.com/cocos-creator/cocos2d-x-lite/pull/3161

PR #3150, opened 2020-12-26
[3.0] cmake: combine ios simulator & device

PatriceJiang:
合并 iOS 的真机和模拟器 到同一个工程

ref: https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#switching-between-device-and-simulator


PatriceJiang:
@minggo  麻烦review 下

PR #3174, opened 2020-12-30
GLES2 texture issue

hana-alice:
texture disappear in gles2


hana-alice:
https://github.com/cocos-creator/3d-tasks/issues/5745

minggo (approved):

PR #3183, opened 2021-01-06
fix crash when pass -1.0f to clearDepth in Metal

hana-alice:
https://github.com/cocos-creator/3d-tasks/issues/5799

validated both on ios and mac.


hana-alice:
@minggo tested, have a review on this.

minggo:
I think should map [-1, 1] into [0, 1]

PR #3184, opened 2021-01-06
[3.0] Android: reserve JNI function symbols

PatriceJiang:
- 简化编译配置
- android linker: `lld`


PatriceJiang:
@minggo 麻烦review下

PatriceJiang:
https://github.com/cocos-creator/cocos2d-x-lite/pull/3194 包含修改, 关闭

PR #3185, opened 2021-01-07
fix clearDepth crash in Metal

hana-alice:
now map depth [-1.0f, 1.0f] to [0.0f, 1.0f] in metal and clamped.


hana-alice:
fix https://github.com/cocos-creator/3d-tasks/issues/5799

minggo:
I think add assert is better than clamf, the user should make sure pass valid data.

hana-alice:
> I think add assert is better than clamf, the user should make sure pass valid data.

image
image

captured by NSight, actually there is a clamp inside opengl functions.

PR #3281, opened 2021-02-05
add support for switching to full screen.

xianyinchen:
manual change the code, app will start into fullscreen.

win32:
build/jsb-link/frameworks/runtime-src/proj.win32/main.cpp
ALT+F4 to force close window.

macOS:
/Users/chenxianyin/Downloads/demo/build/jsb-link/frameworks/runtime-src/proj.ios_mac/mac/main.cpp
command+Q to force close window.

```
AppDelegate app(MACOS_WIN_SIZE_WIDTH, MACOS_WIN_SIZE_HEIGHT, true);
```

### DIFF
--- a/cocos/platform/CCApplication.h
+++ b/cocos/platform/CCApplication.h
@@ -101,7 +101,7 @@ public:
     // This class is useful for internal usage.
     static Application* getInstance() { return _instance; }
     
-    Application(const std::string& name, int width, int height);
+    Application(const std::string& name, int width, int height, bool fullscreen);
     virtual ~Application();
     
     virtual bool applicationDidFinishLaunching();
@@ -190,7 +190,7 @@ protected:
     virtual void onCreateView(PixelFormat& pixelformat, DepthFormat& depthFormat, int& multisamplingCount);
     
 private:
-    void createView(const std::string& name, int width, int height);
+    void createView(const std::string& name, int& width, int& height, bool fullscreen);
     
     static Application* _instance;
     static std::shared_ptr<Scheduler> _scheduler;
--- a/cocos/platform/CCApplication.h
+++ b/cocos/platform/CCApplication.h
@@ -101,7 +101,7 @@ public:
     // This class is useful for internal usage.
     static Application* getInstance() { return _instance; }
     
-    Application(const std::string& name, int width, int height, bool fullscreen);
+    Application(const std::string& name, int width, int height, bool fullscreen = false);
     virtual ~Application();
     
     virtual bool applicationDidFinishLaunching();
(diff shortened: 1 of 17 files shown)


cocos/platform/CCApplication.h
```diff
@@ -190,7 +190,7 @@ class CC_DLL Application
     virtual void onCreateView(PixelFormat& pixelformat, DepthFormat& depthFormat, int& multisamplingCount);
     
 private:
-    void createView(const std::string& name, int width, int height);
+    void createView(const std::string& name, int& width, int& height, bool fullscreen);
```
minggo, line 193: why use reference for int?

cocos/platform/CCApplication.h
```diff
@@ -101,7 +101,7 @@ class CC_DLL Application
     // This class is useful for internal usage.
     static Application* getInstance() { return _instance; }
     
-    Application(const std::string& name, int width, int height);
+    Application(const std::string& name, int width, int height, bool fullscreen);
```
minggo, line 104: It will break API. I think use default value is better.

cocos/platform/CCApplication.h
xianyinchen: This modification is tricky because the full screen is controlled by the bottom layer and the top layer needs to know the modified screen size, so it uses a reference to pass back the width and height of the full screen.

cocos/platform/CCApplication.h
xianyinchen: Yes, already added.

cocos/platform/CCApplication.h
PatriceJiang: 感觉没有必要通过引用修改. 后续代码应该通过 `getSize` 方法获取 宽高.

cocos/platform/win32/CCApplication-win32.cpp
```diff
@@ -101,12 +101,12 @@ NS_CC_BEGIN
...
 std::shared_ptr<Scheduler> Application::_scheduler = nullptr;
 
-Application::Application(const std::string& name, int width, int height)
+Application::Application(const std::string& name, int width, int height, bool fullscreen)
 {
     Application::_instance = this;
     _scheduler = std::make_shared<Scheduler>();
 
-    createView(name, width, height);
+    createView(name, width, height, fullscreen);
```
PatriceJiang, line 109: `fullscreen` 的值应存储到  Application 中, 可以避免来回传, 未来也可以通过提供 `isFullscreen()` 之类的方法实现其他业务.

xianyinchen: 好的，我再优化下回传的情况，可以补充下接口

PR #3285, opened 2021-02-06
fix: do not draw the buffer object if glBufferData failed

wzpan:
If `glBufferData()` failed, it's easily to get SIGSEGV crash when calling `glDrawElements()`.

This PR add a `glGetError()` checking after calling `glBufferData()` to avoid calling `glDrawElements()` after `glBufferData()` failed.

Related discussion: https://forum.cocos.org/t/topic/104524/6

### DIFF
--- a/cocos/renderer/gfx/DeviceGraphics.cpp
+++ b/cocos/renderer/gfx/DeviceGraphics.cpp
@@ -409,6 +409,15 @@ void DeviceGraphics::draw(size_t base, GLsizei count)
     }
     
     commitTextures();
+
+    if(!nextIndexBuffer->isDrawable()){
+        _drawCalls++;
+        auto temp = _nextState;
+        _nextState = _currentState;
+        _currentState = temp;
+        _nextState->reset();
+        return;
+    }
     
     //commit uniforms
     const auto& uniformsInfo = _nextState->getProgram()->getUniforms();
--- a/cocos/renderer/gfx/DeviceGraphics.cpp
+++ b/cocos/renderer/gfx/DeviceGraphics.cpp
@@ -389,6 +389,14 @@ void DeviceGraphics::draw(size_t base, GLsizei count)
     commitVertexBuffer();
     
     auto nextIndexBuffer = _nextState->getIndexBuffer();
+
+    if(!nextIndexBuffer->isDrawable()){
+        _drawCalls++;
+        std::swap(_nextState, _currentState);
+        _nextState->reset();
+        return;
+    }
+
     if (_currentState->getIndexBuffer() != nextIndexBuffer)
     {
         GL_CHECK(ccBindBuffer(GL_ELEMENT_ARRAY_BUFFER, nextIndexBuffer ? nextIndexBuffer->getHandle() : 0));
@@ -409,15 +417,6 @@ void DeviceGraphics::draw(size_t base, GLsizei count)
     }
     
     commitTextures();
-
-    if(!nextIndexBuffer->isDrawable()){
-        _drawCalls++;
-        auto temp = _nextState;
-        _nextState = _currentState;
-        _currentState = temp;
-        _nextState->reset();
-        return;
-    }
     
     //commit uniforms
     const auto& uniformsInfo = _nextState->getProgram()->getUniforms();
--- a/cocos/renderer/gfx/IndexBuffer.cpp
+++ b/cocos/renderer/gfx/IndexBuffer.cpp
@@ -50,6 +50,7 @@ bool IndexBuffer::init(DeviceGraphics* device, IndexFormat format, Usage usage, 
     _usage = usage;
     _numIndices = numIndices;
     _bytesPerIndex = 0;
+    _drawable = true;
     
     _needExpandDataStore = true;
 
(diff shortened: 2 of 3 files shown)


cocos/renderer/gfx/IndexBuffer.cpp
```diff
@@ -110,11 +111,25 @@ void IndexBuffer::update(uint32_t offset, const void* data, size_t dataByteLengt
...
     else
     {
         glBufferSubData(GL_ELEMENT_ARRAY_BUFFER, (GLintptr)offset, (GLsizeiptr)dataByteLength, (const GLvoid*)data);
+        GLenum gl_err = glGetError();
+        if (0 != gl_err){
+            RENDERER_LOGE("GL error 0x%x: %s:%s", gl_err, glEnumName(gl_err), __FUNCTION__);
+            _drawable = false;
+        }else{
+            _drawable = true;
+        }
```
minggo, line 132: Duplicate codes. I know may want to log the function line, but can use `_needExpandDataStore ` to know which codes are executed.

cocos/renderer/gfx/IndexBuffer.cpp
```diff
@@ -110,11 +111,25 @@ void IndexBuffer::update(uint32_t offset, const void* data, size_t dataByteLengt
...
         _needExpandDataStore = false;
     }
     else
     {
         glBufferSubData(GL_ELEMENT_ARRAY_BUFFER, (GLintptr)offset, (GLsizeiptr)dataByteLength, (const GLvoid*)data);
+        GLenum gl_err = glGetError();
```
minggo, line 126: Please use glError instead to obey the coding style.

cocos/renderer/gfx/DeviceGraphics.cpp
```diff
@@ -409,6 +409,15 @@ void DeviceGraphics::draw(size_t base, GLsizei count)
     }
     
     commitTextures();
+
+    if(!nextIndexBuffer->isDrawable()){
+        _drawCalls++;
+        auto temp = _nextState;
+        _nextState = _currentState;
+        _currentState = temp;
```
minggo, line 417: use std::swap() is better

cocos/renderer/gfx/DeviceGraphics.cpp
```diff
@@ -409,6 +409,15 @@ void DeviceGraphics::draw(size_t base, GLsizei count)
     }
     
     commitTextures();
+
+    if(!nextIndexBuffer->isDrawable()){
```
minggo, line 413: May be can check it as early as possible. For example, add after `auto nextIndexBuffer = _nextState->getIndexBuffer();`.

cocos/renderer/gfx/IndexBuffer.cpp
```diff
@@ -110,11 +111,25 @@ void IndexBuffer::update(uint32_t offset, const void* data, size_t dataByteLengt
     {
 //        glBufferData(GL_ELEMENT_ARRAY_BUFFER, (GLsizeiptr)dataByteLength, data, glUsage);
         glBufferData(GL_ELEMENT_ARRAY_BUFFER, _bytes, (const GLvoid*)data, glUsage);
+        GLenum gl_err = glGetError();
+        if (0 != gl_err){
+            RENDERER_LOGE("GL error 0x%x: %s:%s", gl_err, glEnumName(gl_err), __FUNCTION__);
+            _drawable = false;
```
minggo, line 117: Should initialize it in `init()`.

minggo: As the object will be destroyed and initialized again.

cocos/renderer/gfx/IndexBuffer.cpp
wzpan: Actually both blocks will set `_needExpandDataStore` as false after execution. 
But it's fine to catch them all in one piece of code. I will modify it a little bit.

cocos/renderer/gfx/IndexBuffer.cpp
wzpan: You meant init `gl_err` object? I can't find any example code doing that. How?

minggo: Sorry, i mean `_drawable`.

minggo:
I think it is better to check gl error in debug mode, as it is time consuming.

cocos/renderer/gfx/IndexBuffer.cpp
wzpan: I see. Gotcha.

wzpan:
> I think it is better to check gl error in debug mode, as it is time consuming.

But if so how can we catch the crash in release version?

minggo:
> > I think it is better to check gl error in debug mode, as it is time consuming.
> 
> But if so how can we catch the crash in release version?

The invoker should check the data before updating.

minggo:
> If glBufferData() failed, it's easily to get SIGSEGV crash when calling glDrawElements().

What kind of error of glBufferData() will cause SIGSEGV?

wzpan:
> > If glBufferData() failed, it's easily to get SIGSEGV crash when calling glDrawElements().
> 
> What kind of error of glBufferData() will cause SIGSEGV?

As you can see in the topic: https://forum.cocos.org/t/topic/104524/6

`glBufferData()` itself won't cause SIGSEGV, but `glDrawElements()` do when trying to draw bad buffer objects. We need a way to spot the case.

minggo:
> but glDrawElements() do when trying to draw bad buffer objects

I can not find crash information in [the doc](https://docs.gl/es2/glDrawElements). Could you please provide an example to reproduce it?

wzpan:
> > but glDrawElements() do when trying to draw bad buffer objects
> 
> I can not find crash information in [the doc](https://docs.gl/es2/glDrawElements). Could you please provide an example to reproduce it?

Sadly the crash happens occasionally and we also cannot find a reproduceable example. However if you revert the PR #2884 and try the demo of #2930 in iOS you will get a similar crash.

wzpan:
@minggo I've submitted the change according to the review option. But I have to say that I do this at my home since it's holiday. Hence I haven't fully test it. Please test them further if you decide to merge it.

minggo:
> > > but glDrawElements() do when trying to draw bad buffer objects
> > 
> > 
> > I can not find crash information in [the doc](https://docs.gl/es2/glDrawElements). Could you please provide an example to reproduce it?
> 
> Sadly the crash happens occasionally and we either cannot find a reproduceable example. However if you revert the PR #2884 and try the demo of #2930 in iOS you will get a similar crash.

I can understand the modification of #2884, as passed in data is invalid. Except that, i do not know why it will cause crash in draw elements. If passed in pointer doesn't contain enough data, then i think we should check the passed data in binding codes.

minggo:
> @minggo I've submitted the change according to the review option. But I have to say that I do this at my home since it's holiday. Hence I haven't fully test it. Please test them further if you decide to merge it.

Thanks.

wzpan:
> > > > but glDrawElements() do when trying to draw bad buffer objects
> > > 
> > > 
> > > I can not find crash information in [the doc](https://docs.gl/es2/glDrawElements). Could you please provide an example to reproduce it?
> > 
> > 
> > Sadly the crash happens occasionally and we either cannot find a reproduceable example. However if you revert the PR #2884 and try the demo of #2930 in iOS you will get a similar crash.
> 
> I can understand the modification of #2884, as passed in data is invalid. Except that, i do not know why it will cause crash in draw elements. If passed in pointer doesn't contain enough data, then i think we should check the passed data in binding codes.

I agree. Do the checking of passed data is prefered for performance concern.

PR #3288, opened 2021-02-19
avoid default NDK download after delete :instantapp project

xianyinchen:
because include a jniLibs.srcDirs setting, so need add ndkPath, or it will download the default NDK version if it not exist.


wuzhiming:
可能还是要先查下根本的原因，如果按照这样理解，应该不管删不删instant模块都会去下载猜度。

xianyinchen:
目前还没有定位到根本原因，规避方法是在 libcocos和libservice添加ndkPath PROP_NDK_PATH，我这边继续跟踪，PR先关闭了

PR #3294, opened 2021-03-02
Fix the material and resources of the deferred pipeline.

caryliu1999:


cocos/base/threading/ThreadPool.h
```diff
@@ -46,7 +46,7 @@ class ThreadPool final {
     static uint8_t const CPU_CORE_COUNT;
     static uint8_t const MAX_THREAD_COUNT;
 
-    ThreadPool() noexcept          = default;
+    ThreadPool()          = default;
```
minggo, line 49: code format

PR #3297, opened 2021-03-02
support  gen-simulator-release gulp task

PPpro:
for: https://forum.cocos.org/t/topic/106721

changeLog:
- `gen-simulator` task is for user to customize the native simulator project
- `gen-simulator-release` task is for publishing Cocos Creator that  would clean the simulator project after compiling
- remove `make-simulator` task


PPpro:
@minggo pls review

PR #3303, opened 2021-03-09
basic framegraph

YunHsiao:
* resource management
* pass culling & merging 
* load/store op deduction
* graphviz visualization

### DIFF
--- a/cocos/renderer/frame-graph/Blackboard.h
+++ b/cocos/renderer/frame-graph/Blackboard.h
@@ -1,0 +1,85 @@
+/****************************************************************************
+ Copyright (c) 2021 Xiamen Yaji Software Co., Ltd.
+
+ http://www.cocos.com
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated engine source code (the "Software"), a limited,
+ worldwide, royalty-free, non-assignable, revocable and non-exclusive license
+ to use Cocos Creator solely to develop games on your target platforms. You shall
+ not use Cocos Creator software for developing other software or tools that's
+ used for developing games. You are not granted to publish, distribute,
+ sublicense, and/or sell copies of Cocos Creator.
+
+ The software or tools in this License Agreement are licensed, not sold.
+ Xiamen Yaji Software Co., Ltd. reserves all rights not expressly granted to you.
+
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ THE SOFTWARE.
+****************************************************************************/
+
+#pragma once
+
+#include "Handle.h"
+#include "base/Macros.h"
+
+namespace cc {
+namespace framegraph {
+
+template <typename KeyType, typename ValueType, ValueType InvalidValue>
+class Blackboard final {
+public:
+    Blackboard() noexcept          = default;
+    ~Blackboard()                  = default;
+    Blackboard(const Blackboard &) = delete;
+    Blackboard(Blackboard &&)      = delete;
+    Blackboard &operator=(const Blackboard &) = delete;
+    Blackboard &operator=(Blackboard &&) = delete;
+
+    CC_INLINE ValueType &operator[](const KeyType &name) noexcept;
+    CC_INLINE void       put(const KeyType &name, ValueType const handle) noexcept;
+    CC_INLINE ValueType  get(const KeyType &name) const noexcept;
+    CC_INLINE void       clear() noexcept;
+    CC_INLINE bool       has(const KeyType &name) const noexcept;
+
+private:
+    using Container = std::map<KeyType, ValueType>;
+    Container _container{};
+};
+
+//////////////////////////////////////////////////////////////////////////
+
+template <typename KeyType, typename ValueType, ValueType InvalidValue>
+ValueType &Blackboard<KeyType, ValueType, InvalidValue>::operator[](const KeyType &name) noexcept {
+    const auto it = _container.find(name);
+    return (it == _container.end() ? _container.emplace(name, InvalidValue).first : it)->second;
+}
+
+template <typename KeyType, typename ValueType, ValueType InvalidValue>
+void Blackboard<KeyType, ValueType, InvalidValue>::put(const KeyType &name, ValueType const handle) noexcept {
+    _container[name] = handle;
+}
+
+template <typename KeyType, typename ValueType, ValueType InvalidValue>
+ValueType Blackboard<KeyType, ValueType, InvalidValue>::get(const KeyType &name) const noexcept {
+    const auto it = _container.find(name);
+    return it == _container.end() ? InvalidValue : it->second;
+}
+
+template <typename KeyType, typename ValueType, ValueType InvalidValue>
+void Blackboard<KeyType, ValueType, InvalidValue>::clear() noexcept {
+    _container.clear();
+}
+
+template <typename KeyType, typename ValueType, ValueType InvalidValue>
+bool Blackboard<KeyType, ValueType, InvalidValue>::has(const KeyType &name) const noexcept {
+    return _container.find(name) != _container.end();
+}
+
+} // namespace framegraph
+} // namespace cc
--- a/cocos/renderer/frame-graph/CallbackPass.h
+++ b/cocos/renderer/frame-graph/CallbackPass.h
@@ -1,0 +1,72 @@
+/****************************************************************************
+ Copyright (c) 2021 Xiamen Yaji Software Co., Ltd.
+
+ http://www.cocos.com
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated engine source code (the "Software"), a limited,
+ worldwide, royalty-free, non-assignable, revocable and non-exclusive license
+ to use Cocos Creator solely to develop games on your target platforms. You shall
+ not use Cocos Creator software for developing other software or tools that's
+ used for developing games. You are not granted to publish, distribute,
+ sublicense, and/or sell copies of Cocos Creator.
+
+ The software or tools in this License Agreement are licensed, not sold.
+ Xiamen Yaji Software Co., Ltd. reserves all rights not expressly granted to you.
+
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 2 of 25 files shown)


minggo:
I don't think we should implement frame graph in v3.1. V3.1 is mainly for multiple thread rendering and deferred rendering.

YunHsiao:
I think it's okay to put these here as we could restrain from using them in the pipeline for now. This way we could save lots of branching syncing work, and be more flexible on the refactoring work that involves all the modules.

minggo:
Will it affect current logic if not use it?

YunHsiao:
No, except for the public interface changes if it comes to that, but for now it's fine.

cocos/renderer/frame-graph/Blackboard.h
```diff
@@ -0,0 +1,85 @@
...
+template <typename KeyType, typename ValueType, ValueType InvalidValue>
+ValueType &Blackboard<KeyType, ValueType, InvalidValue>::operator[](KeyType const &name) noexcept {
+    auto const it = _container.find(name);
+    return (it == _container.end() ? _container.emplace(name, InvalidValue).first : it)->second;
+}
+
+template <typename KeyType, typename ValueType, ValueType InvalidValue>
+void Blackboard<KeyType, ValueType, InvalidValue>::put(KeyType const &name, ValueType const handle) noexcept {
+    _container[name] = handle;
+    ;
```
minggo, line 65: unneeded code

cocos/renderer/frame-graph/Blackboard.h
```diff
@@ -0,0 +1,85 @@
...
+}
+
+template <typename KeyType, typename ValueType, ValueType InvalidValue>
+void Blackboard<KeyType, ValueType, InvalidValue>::clear() noexcept {
+    _container.clear();
+}
+
+template <typename KeyType, typename ValueType, ValueType InvalidValue>
+bool Blackboard<KeyType, ValueType, InvalidValue>::has(KeyType const &name) const noexcept {
+    return _container.find(name) == _container.end();
```
minggo, line 81: `==` -> `!=`?

cocos/renderer/frame-graph/CallbackPass.h
```diff
@@ -0,0 +1,72 @@
...
+
+private:
+    Data          _data;
+    ExecuteMethod _execute;
+};
+
+//////////////////////////////////////////////////////////////////////////
+
+template <typename Data, typename ExecuteMethod>
+CallbackPass<Data, ExecuteMethod>::CallbackPass(ExecuteMethod &&execute) noexcept
```
minggo, line 61: execute should be a functional, right? If so, then it is better to guard it using `std::is_function`.

cocos/renderer/frame-graph/DevicePass.cpp
```diff
@@ -0,0 +1,186 @@
...
+        append(graph, passNode, attachments);
+    }
+
+    std::sort(attachments.begin(), attachments.end(), RenderTargetAttachment::Sorter());
+    std::vector<gfx::GFXObject const *> renderTargets;
+
+    for (RenderTargetAttachment const &attachment : attachments) {
+        _attachments.emplace_back();
+        Attachment &deviceAttachment     = _attachments.back();
+        deviceAttachment.attachment     = std::move(attachment);
```
minggo, line 50: std::move has not effect on const reference variables.

cocos/renderer/frame-graph/DevicePass.cpp
```diff
@@ -0,0 +1,186 @@
...
+        ResourceNode const &resourceNode = graph.getResourceNode(attachment.textureHandle);
+        assert(resourceNode.virtualResource);
+        deviceAttachment.renderTarget = (gfx::Texture *)resourceNode.virtualResource->getDeviceResource();
```
minggo, line 53: Please do not use C style cast.

cocos/renderer/frame-graph/DevicePassResourceTable.h
```diff
@@ -0,0 +1,64 @@
...
+    DevicePassResourceTable &operator=(DevicePassResourceTable &&) = delete;
+
+    gfx::GFXObject *getRead(Handle const handle) const noexcept;
+    gfx::GFXObject *getWrite(Handle const handle) const noexcept;
+
+private:
+    using ResourceDictionary = std::unordered_map<Handle, gfx::GFXObject *, Handle::Hasher>;
+
+    static gfx::GFXObject *get(ResourceDictionary const &from, Handle const handle) noexcept;
+    void                   extract(FrameGraph const &graph, PassNode const *const passNode, bool const multiSubPass, std::vector<gfx::GFXObject const *> const &renderTargets) noexcept;
```
minggo, line 54: Is const needed for basic types, such as bool, int and so on?

cocos/renderer/frame-graph/FrameGraph.cpp
```diff
@@ -0,0 +1,622 @@
...
+#include "PassNodeBuilder.h"
+#include "Resource.h"
+#include <algorithm>
+#include <fstream>
+#include <set>
+
+namespace cc {
+namespace framegraph {
+
+static StringPool sStringPool;
```
minggo, line 37: i think anonymous is better.

cocos/renderer/frame-graph/FrameGraph.h
```diff
@@ -0,0 +1,145 @@
...
+
+    FrameGraph()                   = default;
+    ~FrameGraph();
+    FrameGraph(FrameGraph const &) = delete;
+    FrameGraph(FrameGraph &&)      = delete;
+    FrameGraph &operator=(FrameGraph const &) = delete;
+    FrameGraph &operator=(FrameGraph &&) = delete;
+
+    static StringHandle stringToHandle(char const *const name) noexcept;
+    static char const * handleToString(StringHandle const handle) noexcept;
```
minggo, line 54: Why use char* instead of String?

YunHsiao: for the names are almost always literals, and the string pool maintains them as chunks of raw memory.

minggo: But if it is passed by `string.c_str()`, then it is strange.

cocos/renderer/frame-graph/VirtualResource.h
```diff
@@ -0,0 +1,74 @@
...
+
+private:
+    PassNode *         _firstUsePass{nullptr};
+    PassNode *         _lastUsePass{nullptr};
+    StringHandle const _name;
+    uint32_t           _refCount{0};
+    uint16_t           _writerCount{0};
+    ID                 _id{0};
+    uint8_t            _version{0};
+    bool const         _imported{false};
```
minggo, line 64: Is there any reason to use const?

YunHsiao: it's constant throughout the object's lifecycle

cocos/renderer/frame-graph/ResourceEntry.h
```diff
@@ -0,0 +1,85 @@
...
+    ResourceEntry()                      = delete;
+    ~ResourceEntry()                     = default;
+    ResourceEntry(ResourceEntry const &) = delete;
+    ResourceEntry(ResourceEntry &&)      = delete;
+    ResourceEntry &operator=(ResourceEntry const &) = delete;
+    ResourceEntry &operator=(ResourceEntry &&) = delete;
+
+    virtual void            request() noexcept override;
+    virtual void            release() noexcept override;
+    virtual gfx::GFXObject *getDeviceResource() const noexcept override;
```
minggo, line 47: virtual is not needed.

cocos/renderer/frame-graph/ResourceEntry.h
```diff
@@ -0,0 +1,85 @@
...
+
+    inline ResourceType &get() noexcept { return _resource; }
```
minggo, line 49: Change to `inline ResourceType &get() const noexcept`?

cocos/renderer/frame-graph/Handle.h
```diff
@@ -0,0 +1,46 @@
...
+using Handle       = cc::IndexHandle<uint16_t>;
+using StringHandle = cc::StringHandle;
+using StringPool   = cc::StringPool<false>;
+
+template <typename T>
+class TypedHandle final : public Handle {
+public:
+    using Type = T;
+    using Handle::Handle;
+};
```
minggo, line 43: TypedHandle is used just for type distinguish? I can not find any other feature.

YunHsiao: this will be useful for type deduction in various situations, coming soon

cocos/renderer/frame-graph/VirtualResource.h
minggo: Got it, interesting.

cocos/renderer/frame-graph/FrameGraph.h
minggo: Sorry, i mean it is dangerous as the string may be destroyed.

cocos/renderer/frame-graph/CallbackPass.h
minggo: Oh, i thought it does.

cocos/renderer/frame-graph/FrameGraph.h
YunHsiao: the string pool makes an internal copy of every enlisted string, so it is safe here

cocos/renderer/frame-graph/PassNode.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+    CC_ASSERT((_usedRenderTargetSlotMask & (1 << attachment.desc.slot)) == 0);
+    _usedRenderTargetSlotMask |= (1 << attachment.desc.slot);
+
+    _attachments.emplace_back(attachment);
+    _hasClearedAttachment = _hasClearedAttachment || (attachment.desc.loadOp == gfx::LoadOp::CLEAR);
+}
+
+bool PassNode::canMerge(FrameGraph const &graph, PassNode const &passNode) const noexcept {
+    size_t const attachmentCount = _attachments.size();
```
caryliu1999, line 102: Is it need to consider the viewport and subpass of the PassNode here？

cocos/renderer/frame-graph/Blackboard.h
```diff
@@ -0,0 +1,84 @@
...
+class Blackboard final {
+public:
+    Blackboard() noexcept          = default;
+    Blackboard(Blackboard const &) = delete;
+    Blackboard(Blackboard &&)      = delete;
+    ~Blackboard()                  = default;
+    Blackboard &operator=(Blackboard const &) = delete;
+    Blackboard &operator=(Blackboard &&) = delete;
+
+    inline ValueType &operator[](KeyType const &name) noexcept;
```
minggo, line 43: name -> key is better?

cocos/renderer/frame-graph/Resource.h
```diff
@@ -0,0 +1,170 @@
...
+template <typename DeviceResourceType, typename DescriptorType, typename DeviceResourceCreatorType, typename DescriptorHasherType>
+void Resource<DeviceResourceType, DescriptorType, DeviceResourceCreatorType, DescriptorHasherType>::createTransient() noexcept {
+    computeHash();
+    _deviceObject = Allocator::getInstance().alloc(_desc, _hash);
+}
+
+template <typename DeviceResourceType, typename DescriptorType, typename DeviceResourceCreatorType, typename DescriptorHasherType>
+void Resource<DeviceResourceType, DescriptorType, DeviceResourceCreatorType, DescriptorHasherType>::createPersistent() noexcept {
+    computeHash();
+
```
caryliu1999, line 104: createTransient and createPersistent have duplicate hash calculations. In addition, the words Transient and Persistent seem like these two methods should be used as a choice between one or the other, but as I've seen in other engines, they should be used as a pre and a post method in the Resource creation process.

cocos/renderer/frame-graph/Blackboard.h
minggo: name -> key.
It seems it is a key instead of a name.

cocos/renderer/frame-graph/Resource.h
YunHsiao: I think the transient/persist concept here is completely different from what you mean there. 

Transient resources are created/reused in object pools, while persistent resources are owned by dedicated resource instances, and imported as external resources to the graph every frame. So the hash calculation function could be the same since indeed, one resource can only either be transient or persistent.

Could you elaborate on your side too?

caryliu1999 (approved):

PR #3313, opened 2021-03-11
Fix seekTo invalid after stopping mediaPlayer

knoxHuang:
https://github.com/cocos-creator/3d-tasks/issues/6259

### DIFF
--- a/cocos/platform/android/java/src/com/cocos/lib/CocosVideoView.java
+++ b/cocos/platform/android/java/src/com/cocos/lib/CocosVideoView.java
@@ -187,7 +187,6 @@ public class CocosVideoView extends SurfaceView {
         if (! (mCurrentState == State.IDLE ||
                 mCurrentState == State.ERROR ||
                 mCurrentState == State.INITIALIZED ||
-                mCurrentState == State.STOPPED ||
                 mMediaPlayer == null) ) {
             mPosition = mMediaPlayer.getCurrentPosition();
         }
@@ -198,7 +197,6 @@ public class CocosVideoView extends SurfaceView {
         if (! (mCurrentState == State.IDLE ||
                 mCurrentState == State.ERROR ||
                 mCurrentState == State.INITIALIZED ||
-                mCurrentState == State.STOPPED ||
                 mMediaPlayer == null) ) {
             mDuration = mMediaPlayer.getDuration();
         }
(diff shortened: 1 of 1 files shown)


cocos/platform/android/java/src/com/cocos/lib/CocosVideoView.java
```diff
@@ -247,14 +245,9 @@ public void stop() {
         if (!(mCurrentState == State.IDLE || mCurrentState == State.INITIALIZED || mCurrentState == State.ERROR || mCurrentState == State.STOPPED)
                 && mMediaPlayer != null) {
             mCurrentState = State.STOPPED;
-            mMediaPlayer.stop();
+            mMediaPlayer.pause();
```
minggo, line 248: Why pause instead of stop?

knoxHuang: using stop will cause seekTo to be invalid unless the MediaPlayer is reinitialized.

minggo: Then there is not difference between pause and stop.

knoxHuang: There is still a distinction between the upper layer, but the bottom layer uses pause and seekTo to fake the stop behavior, and the web platform is also handled in this way

minggo: `and the web platform is also handled in this way`

Why all platforms need to do like this? Why support `seekto` after it is stopped?

knoxHuang: > Why all platforms need to do like this? Why support seekto after it is stopped?
Not to unify it, This is to fix the problem that seekTo cannot be used again without reinitializing after stopping

minggo: Yep, what i mean is that. Is it needed to support seekTo after it is stopped.

knoxHuang: Example:

video.play();
video.stop();
video.play();
video.seekTo(10); // invalid

video.play();
video.puase();
video.seekTo(10); // success

minggo: Then i think the issue is not stopped, because after stopped, it is prepared again by `mMediaPlayer.prepare();`. If media player is prepared, then it can be seek to.

knoxHuang: image

This is the previous code, but it doesn't work

minggo: You should know why it can not work, not just avoid it.

PR #3317, opened 2021-03-12
fix editbox multiline mode can't end editing

PPpro:
resolve: https://github.com/cocos-creator/3d-tasks/issues/6522

NOTE: this pr  https://github.com/cocos-creator/engine-native/pull/3295 missed processing the text view


PPpro:
@minggo pls review

PR #3323, opened 2021-03-16
[chore] gfx layer adjustments

YunHsiao:
* move common routines to gfx base instead of copy-pasting them in every backend, which is just ^@#$
* remove the device pointer reference inside every gfx resources
* add an empty backend
* the device manager abstraction: take over the device lifecycle management job from js


cocos/renderer/gfx-base/GFXBuffer.h
```diff
@@ -37,11 +37,12 @@ class CC_DLL Buffer : public GFXObject {
 
     static uint computeHash(const BufferInfo &info);
 
-    virtual bool initialize(const BufferInfo &info)     = 0;
-    virtual bool initialize(const BufferViewInfo &info) = 0;
-    virtual void destroy()                              = 0;
-    virtual void resize(uint size)                      = 0;
-    virtual void update(void *buffer, uint size)        = 0;
+    void initialize(const BufferInfo &info);
+    void initialize(const BufferViewInfo &info);
```
minggo, line 41: Why change return type to void?

YunHsiao: When running asynchronously in different threads, there is no way to get any meaningful return values. We should insert more asserts and build a much more strict and versatile independent validation layer instead.

minggo: Sounds reasonable. Then i think we should modify all gfx classes.

cocos/renderer/gfx-agent/GFXDeviceAgent.h
```diff
@@ -42,9 +42,9 @@ constexpr uint MAX_CPU_FRAME_AHEAD = 1u;
 
 class CC_DLL DeviceAgent final : public Agent<Device> {
 public:
-    using Agent::Agent;
-    DeviceAgent(Device *const actor) noexcept
-    : Agent(actor, nullptr) {}
+    static DeviceAgent *getInstance();
+
+    DeviceAgent(Device *device);
```
minggo, line 47: It is not a good idea to have public constructor and use singleton pattern.

minggo (approved):

PR #3326, opened 2021-03-17
fix that after clicking the virtual Home button to hide the App, the Home button is not hidden after reopening the App.

OnlyKoei:
After the restoration of virtual Home button click to hide App, App Home key to re-open not hidden.


OnlyKoei:
@PatriceJiang 帮忙看一下

minggo (approved):

PatriceJiang (approved):

PR #3332, opened 2021-03-18
Fix: crash when calling a function with the long type return value

CloseToZero:
Fix problem: crash when calling a function with args that has a return
value of long type.


minggo:
Good catch, thanks.

minggo (approved):

PatriceJiang (approved):

PR #3337, opened 2021-03-19
gfx: validator

YunHsiao:
now we can finally go crazy about all manners of validations

### DIFF
--- a/cocos/renderer/gfx-agent/BufferAgent.cpp
+++ b/cocos/renderer/gfx-agent/BufferAgent.cpp
@@ -56,7 +56,7 @@ void BufferAgent::doInit(const BufferInfo &info) {
 
 void BufferAgent::doInit(const BufferViewInfo &info) {
     BufferViewInfo actorInfo = info;
-    actorInfo.buffer         = ((BufferAgent *)info.buffer)->getActor();
+    actorInfo.buffer         = static_cast<BufferAgent *>(info.buffer)->getActor();
 
     ENQUEUE_MESSAGE_2(
         DeviceAgent::getInstance()->getMessageQueue(),
--- a/cocos/renderer/gfx-agent/BufferAgent.cpp
+++ b/cocos/renderer/gfx-agent/BufferAgent.cpp
@@ -68,7 +68,7 @@ void BufferAgent::doInit(const BufferViewInfo &info) {
         });
 }
 
-void BufferAgent::doResize(uint size) {
+void BufferAgent::doResize(uint size, uint count) {
     ENQUEUE_MESSAGE_2(
         DeviceAgent::getInstance()->getMessageQueue(),
         BufferResize,
--- a/cocos/renderer/gfx-validator/BufferValidator.h
+++ b/cocos/renderer/gfx-validator/BufferValidator.h
@@ -1,0 +1,51 @@
+/****************************************************************************
+ Copyright (c) 2020-2021 Xiamen Yaji Software Co., Ltd.
+
+ http://www.cocos.com
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated engine source code (the "Software"), a limited,
+ worldwide, royalty-free, non-assignable, revocable and non-exclusive license
+ to use Cocos Creator solely to develop games on your target platforms. You shall
+ not use Cocos Creator software for developing other software or tools that's
+ used for developing games. You are not granted to publish, distribute,
+ sublicense, and/or sell copies of Cocos Creator.
+
+ The software or tools in this License Agreement are licensed, not sold.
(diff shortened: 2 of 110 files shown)


cocos/renderer/gfx-agent/BufferAgent.cpp
```diff
@@ -68,7 +68,7 @@ void BufferAgent::doInit(const BufferViewInfo &info) {
         });
 }
 
-void BufferAgent::doResize(uint size) {
+void BufferAgent::doResize(uint size, uint count) {
```
minggo, line 71: What is the meaning of `count`?

YunHsiao: size / stride, passing here just to save the repetitive calculation on each backends

minggo: Got it. It is an internal function, no problem. But it is better to add some comment of `count`.

cocos/renderer/gfx-validator/BufferValidator.h
```diff
@@ -0,0 +1,51 @@
...
+
+    void update(void *buffer, uint size) override;
+
+protected:
+    void doInit(const BufferInfo &info) override;
+    void doInit(const BufferViewInfo &info) override;
+    void doResize(uint size, uint count) override;
+    void doDestroy() override;
+
+    vector<uint8_t> _buffer;
```
minggo, line 47: Should record it here? Can use actor's buffer? Or it will consume too much memory.

cocos/renderer/gfx-validator/DeviceValidator.cpp
```diff
@@ -0,0 +1,204 @@
...
+void DeviceValidator::present() {
+    _actor->present();
+}
+
+void DeviceValidator::setMultithreaded(bool multithreaded) {
+    _actor->setMultithreaded(multithreaded);
+}
+
+CommandBuffer *DeviceValidator::createCommandBuffer(const CommandBufferInfo &info, bool hasValidator) {
+    CommandBuffer *actor = _actor->createCommandBuffer(info, true);
```
caryliu1999, line 114: I haven't seen the use of the hasValidator parameter.

cocos/renderer/gfx-validator/BufferValidator.h
YunHsiao: The validator doesn't know about the underlying backend's data structure. It has to maintain its own data.

cocos/renderer/gfx-validator/DeviceValidator.cpp
YunHsiao: Oops, that's a mistake. the parameter is for device agent, not device validator.

cocos/renderer/gfx-validator/BufferValidator.h
minggo: Yep, you are right. But does it need to record data? Will the validation layer check its content?

YunHsiao: Most likely not. that's why the actual allocation only happens inside 'if (device->isRecording())' conditions

minggo: Sounds good.

cocos/renderer/gfx-validator/CommandBufferValidator.cpp
```diff
@@ -0,0 +1,228 @@
...
+}
+
+void CommandBufferValidator::doDestroy() {
+    _actor->destroy();
+}
+
+void CommandBufferValidator::begin(RenderPass *renderPass, uint subpass, Framebuffer *framebuffer) {
+    CCASSERT(!_insideRenderPass, "Already inside an render pass?");
+
+    _insideRenderPass = !!renderPass;
```
minggo, line 62: It is not clear to understand the logic.

cocos/renderer/gfx-validator/CommandBufferValidator.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+    RenderPass * renderPassActor  = renderPass ? ((RenderPassValidator *)renderPass)->getActor() : nullptr;
+    Framebuffer *framebufferActor = framebuffer ? ((FramebufferValidator *)framebuffer)->getActor() : nullptr;
```
minggo, line 65: Please do not use c style cast.

cocos/renderer/gfx-validator/CommandBufferValidator.cpp
YunHsiao: Added some comments to that

minggo (approved):

caryliu1999 (approved):

PR #3339, opened 2021-03-23
fix crash issue port to NSView

hana-alice:


minggo:
I think we can only support Metal on macOS now. I will create an issue for it.

PR #3341, opened 2021-03-23
gfx: lifecyle validations

YunHsiao:
All gfx resources are now tracked and leaks can be reported with initialization call stack information.


cocos/renderer/gfx-base/GFXDescriptorSet.cpp
```diff
@@ -40,6 +40,8 @@ DescriptorSet::~DescriptorSet() {
 }
 
 void DescriptorSet::initialize(const DescriptorSetInfo& info) {
+    CCASSERT(info.layout, "Invalid set layout");
```
minggo, line 43: Why not add it in validate layer?

cocos/renderer/gfx-base/GFXShader.h
```diff
@@ -52,7 +52,7 @@ class CC_DLL Shader : public GFXObject {
 
 protected:
     static uint generateShaderID() noexcept {
-        static uint _idGen = 0;
+        static uint _idGen = 10000;
```
minggo, line 55: Any reason to change the value? And why using `10000`?

YunHsiao: no particular reason, changed this just so that we don't have 0 a valid id.

cocos/renderer/gfx-base/GFXDescriptorSet.cpp
YunHsiao: because the base class needs the `info.layout` to be non-null right away

minggo: Ok, got it.

cocos/renderer/gfx-base/GFXShader.h
minggo: Got it. Is the validation of  shader id is checked in codes?

cocos/renderer/gfx-validator/ValidationUtils.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+#define BOOST_STACKTRACE_GNU_SOURCE_NOT_REQUIRED
+#include "boost/stacktrace.hpp"
+
+namespace cc {
+namespace gfx {
+
+String getStacktrace(uint skip, uint maxDepth) {
+    return boost::stacktrace::to_string(boost::stacktrace::stacktrace(skip, maxDepth));
+}
```
minggo, line 36: I think it is better to move it to base/Utils.h.

minggo, line 36: I think it is better to move it to base/Utils.h.

cocos/renderer/gfx-base/GFXShader.h
YunHsiao: no, not for now, but definitely can do

cocos/base/Utils.h
```diff
@@ -53,6 +54,9 @@ utils::nextPOT function is licensed under the same license that is used in Textu
 */
 
 namespace utils {
+
+CC_DLL std::string getStacktrace(uint skip = 0, uint maxDepth = static_cast<uint>(-1));
```
minggo, line 58: It is better to use [UINT_MAX](https://en.cppreference.com/w/cpp/types/climits) instead of `maxDepth = static_cast<uint>(-1)`.

cocos/renderer/gfx-validator/ValidationUtils.h
```diff
@@ -0,0 +1,125 @@
...
+
+    template <typename T, typename = std::enable_if_t<std::is_same<Resource, T>::value>>
+    static std::enable_if_t<!record_stacktrace<T>::value, void> push(T *resource);
+
+    static void erase(Resource *resource);
+    static void checkEmpty();
+
+private:
+    struct ResourceRecord {
+        Resource *resource;
```
minggo, line 68: It is better to have default value.

minggo (approved):

PR #3344, opened 2021-03-24
[3.0.1] fix websocket crash

PatriceJiang:
fix: https://github.com/cocos-creator/3d-tasks/issues/6675


cocos/network/WebSocket-libwebsockets.cpp
```diff
@@ -545,6 +548,7 @@ class WebSocketFrame {
 
 void WebSocketImpl::closeAllConnections() {
     if (__websocketInstances != nullptr) {
+        std::lock_guard<std::mutex> lk(__instanceMutex);
```
PatriceJiang, line 551:
```suggestion
```

PR #3345, opened 2021-03-24
Coding style doc

minggo:


docs/CODING_STYLE.md
```diff
@@ -0,0 +1,1411 @@
...
+
+**Decision:**
+
+<cstdint> defines types like int16_t, uint32_t, int64_t, etc. You should always use those in preference to short, unsigned long long and the like, when you need a guarantee on the size of an integer. Of the C integer types, only int should be used. When appropriate, you are welcome to use standard types like size_t and ptrdiff_t.
+
+We use int very often, for integers we know are not going to be too big, e.g., loop counters. Use plain old int for such things. You should assume that an int is at least 32 bits, but don't assume that it has more than 32 bits. If you need a 64-bit integer type, use int64_t or uint64_t.
+
+For integers we know can be "big", use int64_t.
+
+You should not use the unsigned integer types such as uint32_t, unless there is a valid reason such as representing a bit pattern rather than a number, or you need defined overflow modulo 2^N. In particular, do not use unsigned types to say a number will never be negative. Instead, use assertions for this.
```
YunHsiao, line 1083: `You should not use the unsigned integer types such as uint32_t`
I think we are doing the exact opposite of that

YunHsiao:
The markdown format is kind of hard to follow. Maybe we should just put a link to the original google doc and write only the part we want to change instead?

minggo:
Yep, i have thought it. But i am afraid that the google's coding style may be modified, and the rule may be changed too.

docs/CODING_STYLE.md
minggo: Good catch. I think we should discuss it more and make a decision. If not use `uint32_t`, then we should find which type to use. May be `int` is better.

docs/CODING_STYLE.md
```diff
@@ -0,0 +1,1411 @@
...
+
+Accessors and mutators (get and set functions) may be named like variables. These often correspond to actual member variables, but this is not required. For example, `int count()` and `void setCount(int count)`.
+
+## Namespace Names
+
+Namespace names are all lower-case, with words separated by underscores. Top-level namespace names are based on the project name . Avoid collisions between nested namespaces and well-known top-level namespaces.
+
+## Enumerator Names
+
+Enumerators (for both scoped and unscoped enums) should be named like constants. That is, use kEnumName not ENUM_NAME.
```
YunHsiao, line 1389: use `ENUM_NAME` not `kEnumName`.

docs/CODING_STYLE.md
```diff
@@ -0,0 +1,1411 @@
...
+
+Correct use of header files can make a huge difference to the readability, size and performance of your code.
+
+The following rules will guide you through the various pitfalls of using header files.
+
+## The #pragma Guard
+
+All header files should have `#pragma once` guards to prevent multiple inclusion.
+
+To guarantee uniqueness, they should be based on the full path in a project's source tree. For example, the file `engine-native/cocos/math/Math.h` in project foo should have the following guard:
```
YunHsiao, line 89: > To guarantee uniqueness, they should be based on the full path in a project's source tree. For example, the file `engine-native/cocos/math/Math.h` in project foo should have the following guard:

redundant

docs/CODING_STYLE.md
minggo: fixed

docs/CODING_STYLE.md
minggo: fixed

docs/CODING_STYLE.md
```diff
@@ -948,11 +946,11 @@ The C++-style cast syntax is verbose and cumbersome.
...
 - Use static_cast as the equivalent of a C-style cast that does value conversion, when you need to explicitly up-cast a pointer from a class to its superclass, or when you need to explicitly cast a pointer from a superclass to a subclass. In this last case, you must be sure your object is actually an instance of the subclass.
 - Use const_cast to remove the const qualifier (see const).
-- Use reinterpret_cast to do unsafe conversions of pointer types to and from integer and other pointer types, including void*. Use this only if you know what you are doing and you understand the aliasing issues. Also, consider the alternative absl::bit_cast.
-- Use absl::bit_cast to interpret the raw bits of a value using a different type of the same size (a type pun), such as interpreting the bits of a double as int64.
+- Use reinterpret_cast to do unsafe conversions of pointer types to and from integer and other pointer types, including void*. Use this only if you know what you are doing and you understand the aliasing issues. Also, consider the alternative `absl::bit_cast`.
```
YunHsiao, line 953: There are many references to this `absl` library which we currently do not have access to. Do we need to import this library?

minggo: No, there is not plan currently. I will remove them.

PR #3350, opened 2021-03-25
remove Array.cpp/.h

minggo:


cocos/base/Scheduler.cpp
```diff
@@ -177,39 +151,36 @@ void Scheduler::schedule(const ccSchedulerFunc &callback, void *target, float in
...
         }
-        ccArrayEnsureExtraCapacity(element->timers, 1);
     }
 
-    TimerTargetCallback *timer = new (std::nothrow) TimerTargetCallback();
+    auto *timer = new (std::nothrow) TimerTargetCallback();
     timer->initWithCallback(this, callback, target, key, interval, repeat, delay);
-    ccArrayAppendObject(element->timers, timer);
+    element->timers.push_back(timer);
     timer->release();
```
PatriceJiang, line 184: `timer` should not be released here, since it can nolonger be retained in `ccArrayAppendObject`

cocos/base/Scheduler.cpp
```diff
@@ -177,39 +151,36 @@ void Scheduler::schedule(const ccSchedulerFunc &callback, void *target, float in
...
         // Is this the 1st element ? Then set the pause level to all the selectors of this target
         element->paused = paused;
     } else {
         CCASSERT(element->paused == paused, "element's paused should be paused!");
     }
 
-    if (element->timers == nullptr) {
-        element->timers = ccArrayNew(10);
+    if (element->timers.empty()) {
+        element->timers.reserve(INITIAL_TIMER_COUND);
```
PatriceJiang, line 169: Can this be omitted, since it is still empty after `reserve`?

PatriceJiang (approved):

cocos/base/Scheduler.cpp
minggo: reserve just make sure to have enough memory, do not have to expand every time. Omit it will not cause error, but here just keep the same logic as before.

cocos/base/Scheduler.cpp
minggo: Yep, you are right.

PR #3363, opened 2021-03-30
[Modify] change the combineSigny initialization location

GengineJS:


GengineJS:
https://github.com/cocos-creator/engine/pull/8406

minggo:
Why need to change it like this?

GengineJS:
> Why need to change it like this?

Mainly in the js engine, the initialize function may not be executed after running. In order to be unified with jsEngine

minggo:
> the initialize function may not be executed after running

@caryliu1999 i think we should fix it. Initialize should be invoked before being used.

GengineJS:
> > the initialize function may not be executed after running
> 
> @caryliu1999 i think we should fix it. Initialize should be invoked before being used.

But to get combineSignY, the device needs to be defined, and the device is assigned in the active function.

minggo:
> But to get combineSignY, the device needs to be defined, and the device is assigned in the active function.

Yep. active() is invoked before initialize(), so device is assigned when invoke initialize().

PR #3373, opened 2021-04-01
fix simulator compiling

PPpro:


PPpro:
@PatriceJiang @minggo pls review

PatriceJiang (approved):

tools/simulator/frameworks/runtime-src/CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ if(CMAKE_OSX_SYSROOT STREQUAL "iphonesimulator")
 else()
     set(TARGET_IOS_VERSION "8.0")
 endif()
-set(TARGET_OSX_VERSION "10.10")
```
minggo, line 34: Why need to change it?

PPpro: can't  use std:get  method when `TARGET_OSX_VERSION` is 10.10

PatriceJiang: 编译 tbb 用到 `std::get`,  要求有 macos 10.14+.    引擎的最低版本也是 10.14.

minggo: 所以还是得去除对 tbb 对依赖

PR #3379, opened 2021-04-04
Clang-Tidy adaptations

YunHsiao:
relevant: https://github.com/cocos-creator/engine/pull/8423

If auto-fix doesn't work and reports this error in the console, update to MSVC 2019 (16.9+):
```
...\intrin.h(56,1): error G7973E533: expected function body after function declarator [clang-diagnostic-error]
```
https://github.com/microsoft/STL/issues/1300

Also, the extension doesn't work in a multi-root workspace environment.


pandamicro (approved):

pandamicro:
Clang tidy seems very strict, how do we apply the requirement ?

minggo:
> Clang tidy seems very strict, how do we apply the requirement ? @minggo

It is caused by add `readability-*`. But if removing it, then readability-identifier-naming.XXX can not take effect. I do not know how to fix it, so we can just ignore the rules that are strict, such as add

```
Checks: >
- xxx,
-readability-uppercase-literal-suffix,
-readability-braces-around-statements
```

minggo:
After some experiment, i think replace `readability-*` with `readability-identifier-*` is better.

minggo:
I have changed to use readability-identifier-*. Please fix the coding style issue. If think some rule are not needed, then we can discuss it.

.clang-tidy
```diff
@@ -1,81 +1,60 @@
...
+    - { key: readability-identifier-naming.ClassConstantCase,      value: UPPER_CASE }
+    - { key: readability-identifier-naming.ClassMemberCase,        value: camelBack  }
+    - { key: readability-identifier-naming.ProtectedMemberPrefix,  value: _          }
+    - { key: readability-identifier-naming.PrivateMemberPrefix,    value: _          }
+
+    - { key: readability-identifier-naming.EnumCase,               value: CamelCase  }
+    - { key: readability-identifier-naming.ScopedEnumConstantCase, value: UPPER_CASE }
+    - { key: readability-identifier-naming.EnumConstantCase,       value: UPPER_CASE }
 
+    - { key: readability-identifier-naming.FunctionCase,           value: aNy_CasE   }
```
minggo, line 53: Why function name is `aNy_CasE`?

cocos/renderer/gfx-gles2/GLES2PrimaryCommandBuffer.cpp
```diff
@@ -38,15 +38,13 @@
 namespace cc {
 namespace gfx {
 
-GLES2PrimaryCommandBuffer::GLES2PrimaryCommandBuffer()
-: GLES2CommandBuffer() {
-}
+GLES2PrimaryCommandBuffer::GLES2PrimaryCommandBuffer() = default;
```
minggo, line 41: Why not define it in .h?

YunHsiao: https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-equals-default.html

.clang-tidy
YunHsiao: Google uses the loose requirement on this. We have many functions that are not strictly following camelBack style, so I think for now we can hold on to this rule?

minggo: If using this rule, then `Class::Func()` will be valid?

YunHsiao: Yes

minggo: I am agree we can hold on this rule and add some comments for existing codes, but then we can not check new codes.

minggo (approved):

PR #3383, opened 2021-04-07
on screen buffer MSAA EGL configurattion / choose config most suitable

hana-alice:
code based on v3.0.1 
c2f6a87fa2ea8248742a2145955bf2129ef87904
April 2.

### DIFF
--- a/cocos/renderer/gfx-gles2/GLES2Context.cpp
+++ b/cocos/renderer/gfx-gles2/GLES2Context.cpp
@@ -148,7 +148,7 @@ bool GLES2Context::initialize(const ContextInfo &info) {
         // TODO: there are more options if not limit rgba to 8.
         EGLint defaultAttribs[] = {
             EGL_SURFACE_TYPE, EGL_WINDOW_BIT | EGL_PBUFFER_BIT,
-            EGL_RENDERABLE_TYPE, EGL_OPENGL_ES3_BIT_KHR,
+            EGL_RENDERABLE_TYPE, EGL_OPENGL_ES2_BIT,
             // EGL_BUFFER_SIZE, colorBuffSize,
             EGL_BLUE_SIZE, blueSize,
             EGL_GREEN_SIZE, greenSize,
--- a/cocos/renderer/gfx-gles2/GLES2Context.cpp
+++ b/cocos/renderer/gfx-gles2/GLES2Context.cpp
@@ -32,7 +32,7 @@ THE SOFTWARE.
     #include "cocos/bindings/event/EventDispatcher.h"
 #endif
 
-#define FORCE_DISABLE_VALIDATION 1
+#define FORCE_DISABLE_VALIDATION 0
 
 namespace cc {
 namespace gfx {
@@ -485,8 +485,12 @@ bool GLES2Context::MakeCurrent(bool bound) {
 
 #if CC_DEBUG > 0 && !FORCE_DISABLE_VALIDATION && CC_PLATFORM != CC_PLATFORM_MAC_IOS
             GL_CHECK(glEnable(GL_DEBUG_OUTPUT_SYNCHRONOUS_KHR));
-            GL_CHECK(glDebugMessageControlKHR(GL_DONT_CARE, GL_DONT_CARE, GL_DONT_CARE, 0, NULL, GL_TRUE));
-            GL_CHECK(glDebugMessageCallbackKHR(GLES2EGLDebugProc, NULL));
+            if (glDebugMessageControlKHR) {
+                GL_CHECK(glDebugMessageControlKHR(GL_DONT_CARE, GL_DONT_CARE, GL_DONT_CARE, 0, NULL, GL_TRUE));
+            }
+            if (glDebugMessageCallbackKHR) {
+                GL_CHECK(glDebugMessageCallbackKHR(GLES2EGLDebugProc, NULL));
+            }
 #endif
 
             _isInitialized = true;
--- a/cocos/renderer/gfx-gles2/GLES2Context.cpp
+++ b/cocos/renderer/gfx-gles2/GLES2Context.cpp
@@ -142,10 +142,9 @@ bool GLES2Context::initialize(const ContextInfo &info) {
         _colorFmt = Format::RGBA8;
         _depthStencilFmt = Format::D24S8;
 
-        bool msaaEnabled = _device->hasFeature(Feature::MSAA);
+        bool msaaEnabled = info.msaaEnabled;
         EGLint redSize{8}, greenSize{8}, blueSize{8}, alphaSize{8}, depthSize{24}, stencilSize{8}, sampleBufferSize{msaaEnabled ? EGL_DONT_CARE : 0}, sampleSize{msaaEnabled ? EGL_DONT_CARE : 0};
 
-        // TODO: there are more options if not limit rgba to 8.
         EGLint defaultAttribs[] = {
             EGL_SURFACE_TYPE, EGL_WINDOW_BIT | EGL_PBUFFER_BIT,
             EGL_RENDERABLE_TYPE, EGL_OPENGL_ES2_BIT,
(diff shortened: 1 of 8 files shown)


hana-alice:
tested on redmiK20 pro / oppo a37m / win32

cocos/renderer/gfx-gles2/GLES2Device.cpp
```diff
@@ -156,6 +141,21 @@ bool GLES2Device::initialize(const DeviceInfo &info) {
...
     }
 
+    _renderContext = CC_NEW(GLES2Context(this));
+    if (!_renderContext->initialize(ctxInfo)) {
+        destroy();
+        return false;
+    }
+    bindRenderContext(true);
+
+    String extStr = (const char *)glGetString(GL_EXTENSIONS);
```
YunHsiao, line 151: Don't move this here, any extension detection before this will be invalidated.

cocos/renderer/gfx-gles3/GLES3Context.cpp
```diff
@@ -137,18 +137,106 @@ bool GLES3Context::initialize(const ContextInfo &info) {
...
+
+        EGLint depth{0}, stencil{0}, sampleBuffers{0}, sampleCount{0};
+
+        // user defined, how muny attrs you want to filter by yourself.
+        const uint8_t attrNums = 8;
+        int64_t lastScore{0};
+        int params[attrNums] = {0};
+
+        const Performance performanceStrategy = Performance::LOW_POWER;
+        const bool performancePreferred = performanceStrategy == Performance::HIGH_QUALITY;
```
YunHsiao, line 186: Not sure the enum is necessary, isn't a constexpr bool literal will do here?

cocos/renderer/gfx-gles3/GLES3Context.cpp
```diff
@@ -137,18 +137,106 @@ bool GLES3Context::initialize(const ContextInfo &info) {
...
+        /* auto compare = performanceStrategy ==  Performance::HIGH_QUALITY ? (>) : (<) ;*/
+        // egl configurations store in "ascending order": https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglChooseConfig.xhtml
+        // so equal applies to >(=)
+        std::function<bool(int, int)> compare;
```
YunHsiao, line 190: std::function introduces overhead, try to avoid it whenever possible. In this case, `(a < b) ^ performancePreferred` will do.

cocos/renderer/gfx-gles3/GLES3Context.cpp
```diff
@@ -137,18 +137,106 @@ bool GLES3Context::initialize(const ContextInfo &info) {
         _colorFmt = Format::RGBA8;
         _depthStencilFmt = Format::D24S8;
 
-        const EGLint attribs[] = {
+        bool msaaEnabled = _device->hasFeature(Feature::MSAA);
+        EGLint redSize{8}, greenSize{8}, blueSize{8}, alphaSize{8}, depthSize{24}, stencilSize{8}, sampleBufferSize{msaaEnabled ? EGL_DONT_CARE : 0}, sampleSize{msaaEnabled ? EGL_DONT_CARE : 0};
+
+        // TODO: there are more options if not limit rgba to 8.
```
YunHsiao, line 143: If you mean float textures, etc. I don't think we need to consider those options. For the purpose of back buffers RGBA8 is perfectly fine.

cocos/renderer/gfx-gles2/GLES2Context.h
```diff
@@ -107,6 +107,7 @@ class CC_GLES2_API GLES2Context final : public Context {
     EGLSurface _eglSurface = EGL_NO_SURFACE;
     EGLContext _eglContext = EGL_NO_CONTEXT;
     EGLContext _eglSharedContext = EGL_NO_CONTEXT;
+    EGLConfig *_eglConfigList = nullptr;
```
YunHsiao, line 110: `std::vector<EGLConfig>` would be better

cocos/renderer/gfx-gles3/GLES3Context.cpp
```diff
@@ -137,18 +137,106 @@ bool GLES3Context::initialize(const ContextInfo &info) {
...
+
+        enum Performance {
+            HIGH_QUALITY,
+            LOW_POWER,
+            //BALANCE,
+        };
+
+        EGLint depth{0}, stencil{0}, sampleBuffers{0}, sampleCount{0};
+
+        // user defined, how muny attrs you want to filter by yourself.
```
YunHsiao, line 180: Note the typos. Also, if we are not exposing this functionality as a user-level modifiable property, it is then not actually 'user-defined', but 'engine-defined'. Please choose the word more carefully and try to convey a more consistent meaning throughout the comments.

cocos/renderer/gfx-gles3/GLES3Context.cpp
```diff
@@ -137,18 +137,106 @@ bool GLES3Context::initialize(const ContextInfo &info) {
...
+        const uint8_t attrNums = 8;
+        int64_t lastScore{0};
```
YunHsiao, line 182: bit-wise masks should use unsigned integer type

cocos/renderer/gfx-gles2/GLES2Context.cpp
```diff
@@ -142,18 +142,106 @@ bool GLES2Context::initialize(const ContextInfo &info) {
         _colorFmt = Format::RGBA8;
         _depthStencilFmt = Format::D24S8;
 
-        const EGLint attribs[] = {
+        bool msaaEnabled = _device->hasFeature(Feature::MSAA);
```
YunHsiao, line 145: This is an incorrect use of the public interface. `hasFeature` serves as a query interface for the upper layer to lookup the feature availabilities. Here the dependency is reversed. I think you need to add a bool property to `DeviceInfo` struct, specifying whether MSAA is desired, and use that here as the request flag. Then after the back buffer initialization processes, return the actual sample count as the evidence for the device feature flag, which the upper layer can now query into.

hana-alice: any docs? this made half work in vain

cocos/renderer/gfx-gles3/GLES3Context.cpp
hana-alice: Just left an example here, performanceStrategy  finally will be a

cocos/renderer/gfx-gles2/GLES2Context.h
hana-alice: just curious why vector is better here, the _eglConfigList size is fixed once created, and seems vector will introduce new std header files in

YunHsiao: https://clang.llvm.org/extra/clang-tidy/checks/modernize-avoid-c-arrays.html

hana-alice: tried vector,
`eglChooseConfig(_eglDisplay, defaultAttribs, _eglConfigList, count, &numConfig)` requires an input-output memory pointer, vector.data() will replace _eglConfigList here but this is not how vector works, cuz we update data with raw pointer hold by this vector without updating its size or capacity, etc. we can do sth hack here make it work, just wondering whether worth it.

YunHsiao: https://www.cplusplus.com/reference/vector/vector/resize/

hana-alice: i was using reserve(size), not saying its feasible or not, i mean this usage is not vector designed for.

YunHsiao: No, it does. Actually, it's quite common to use it this way. And `resize` is the more proper choice than `reserve`, since `reserve` doesn't actually bump the vector size, and it'll introduce problems when using the `.size()` function afterward.

minggo:
@YunHsiao any question about this PR?

YunHsiao (approved):

PR #3386, opened 2021-04-08
[Feature] Complete the math unit test

GengineJS:


unit-test/src/utils.h
```diff
@@ -0,0 +1,19 @@
+//
+//  utils.h
+//  CocosTest
+//
+//  Created by rosa on 2021/3/22.
+//
```
minggo, line 6: Please change the comment.

unit-test/src/utils.h
```diff
@@ -0,0 +1,19 @@
...
+
+#ifndef utils_h
+#define utils_h
```
minggo, line 9: Use `#pragma once` instead.

minggo (approved):

PR #3396, opened 2021-04-13
Fix more warings

mmyduckx:
Related files include win32 and renderer


cocos/renderer/pipeline/RenderAdditiveLightQueue.h
```diff
@@ -64,16 +62,16 @@ class RenderAdditiveLightQueue : public Object {
 private:
     void clear();
     void gatherValidLights(const Camera *camera);
-    bool cullingLight(const Light *light, const ModelView *model);
+    static bool         cullingLight(const Light *light, const ModelView *model);
```
minggo, line 65: Should define static functions first.

minggo (approved):

PR #3397, opened 2021-04-13
fix simulator compiling on Win32

PPpro:
changeLog:
- fix simulator compiling on Win32, reference: https://studiofreya.com/2018/01/06/visual-studio-2017-with-cpp17-and-boost/#stdbyte-ambiguous-symbol-and-rpcndr.h

### DIFF
--- a/tools/simulator/frameworks/runtime-src/Classes/ide-support/RuntimeJsImpl.cpp
+++ b/tools/simulator/frameworks/runtime-src/Classes/ide-support/RuntimeJsImpl.cpp
@@ -30,6 +30,7 @@
 //
 
 // NOTE: fix the conflict typedef 'byte' between C++17 and one Windows header named 'rpcndr.h'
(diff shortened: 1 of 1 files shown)


PPpro:
@minggo pls review

tools/simulator/frameworks/runtime-src/Classes/ide-support/RuntimeJsImpl.cpp
```diff
@@ -29,6 +29,10 @@
 //
 //
 
+#ifdef WIN32
+#define _HAS_STD_BYTE 0
+#endif
```
minggo, line 34: It is better to add some comment to explain it.

PPpro: added

minggo (approved):

PR #3413, opened 2021-04-15
fix lightingQueue memory leak for v3.1

troublemaker52025:


cocos/renderer/pipeline/RenderAdditiveLightQueue.cpp
```diff
@@ -469,5 +450,4 @@ gfx::DescriptorSet *RenderAdditiveLightQueue::getOrCreateDescriptorSet(const Lig
     return _descriptorSetMap.at(light);
 }
 
-} // namespace pipeline
-} // namespace cc
+} // namespace cc::pipeline
```
YunHsiao, line 453: Nested namespace rule is disabled in the config, if you changed this due to linter errors, you might need to double-check your `clang-tidy` environment setup.

troublemaker52025: sure

caryliu1999 (approved):

cocos/renderer/pipeline/RenderAdditiveLightQueue.cpp
```diff
@@ -47,7 +47,7 @@
 namespace cc {
 namespace pipeline {
 
-RenderAdditiveLightQueue::RenderAdditiveLightQueue(RenderPipeline *pipeline) : _pipeline(static_cast<ForwardPipeline *>(pipeline)),
+RenderAdditiveLightQueue::RenderAdditiveLightQueue(RenderPipeline *pipeline) : _pipeline(dynamic_cast<ForwardPipeline *>(pipeline)),
```
minggo, line 50: why use dynamic_cast?

troublemaker52025: When performing downstream conversions (base class rotor subClass), dynamic_cast has type checking and is safer than static_cast.

troublemaker52025: This is what VisualStudio2019 recommends

PR #3414, opened 2021-04-15
Optimize the performance of the memory pool.

caryliu1999:
https://github.com/cocos-creator/3d-tasks/issues/6760


cocos/bindings/dop/BufferAllocator.cpp
```diff
@@ -29,19 +29,21 @@
...
+    }
 }
 
 BufferAllocator::~BufferAllocator() {
     for (auto buffer : _buffers) {
         buffer.second->decRef();
     }
     _buffers.clear();
-    BufferAllocator::_pools.erase(_type);
+    _caches.clear();
```
minggo, line 46: I think _buffers.clear() and  _caches.clear() is not needed in destructor.

cocos/bindings/dop/BufferAllocator.h
```diff
@@ -37,42 +37,58 @@ namespace se {
...
     ~BufferAllocator();
 
     Object *alloc(uint index, uint bytes);
-    void free(uint index);
+    void    free(uint index);
 
 private:
-    static cc::map<PoolType, BufferAllocator *> _pools;
-    static constexpr uint _bufferMask = ~(1 << 30);
+    static cc::vector<BufferAllocator *> _pools;
```
minggo, line 86: Without pool type information, how to find correct pool?

cocos/bindings/dop/BufferAllocator.cpp
caryliu1999: > I think _buffers.clear() and _caches.clear() is not needed in destructor.

Ok, I delete them.

cocos/bindings/dop/BufferAllocator.h
caryliu1999: > Without pool type information, how to find correct pool?

PoolType is continuous, and several different pools are indexed from 0 and then converted to id values based on type.

PR #3419, opened 2021-04-16
onscreen msaa from v3.0.2 to v3.1

hana-alice:


external/config.json
```diff
@@ -1,9 +1,9 @@
 {
-    "version": "v3d-gfx-40",
+    "version": "v3d-gfx-41",
```
minggo, line 2: Sorry, it should be v3d-gfx-42.

minggo:
It seems there is an error in clang tidy. So merge it.

PR #3443, opened 2021-04-21
Optimize the performance of the typed array pool on the native.

caryliu1999:
https://github.com/cocos-creator/3d-tasks/issues/6760

### DIFF
--- a/cocos/bindings/dop/BufferAllocator.cpp
+++ b/cocos/bindings/dop/BufferAllocator.cpp
@@ -43,7 +43,7 @@ BufferAllocator::~BufferAllocator() {
         buffer.second->decRef();
     }
     _buffers.clear();
-    _caches.clear();
+    _array.clear();
 }
 
 Object *BufferAllocator::alloc(uint index, uint bytes) {
@@ -58,7 +58,20 @@ Object *BufferAllocator::alloc(uint index, uint bytes) {
     uint8_t *ret = nullptr;
     size_t   len;
     obj->getArrayBufferData(static_cast<uint8_t **>(&ret), &len);
-    _caches[index] = ret;
+
+    // cache data
+    if (index >= _array.size()) {
+        _array.push_back(ret);
+    } else {
+        _array[index] = ret;
+    }
+
+    // cache size
+    if (index >= _sizes.size()) {
+        _sizes.push_back(bytes);
+    } else {
+        _sizes[index] = bytes;
+    }
 
     return obj;
 }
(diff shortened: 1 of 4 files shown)


cocos/bindings/dop/BufferAllocator.cpp
```diff
@@ -43,7 +43,7 @@ BufferAllocator::~BufferAllocator() {
         buffer.second->decRef();
     }
     _buffers.clear();
-    _caches.clear();
+    _array.clear();
```
minggo, line 46: Do not have to clear them in destructor.

cocos/renderer/pipeline/deferred/DeferredPipeline.cpp
```diff
@@ -128,7 +128,6 @@ bool DeferredPipeline::activate() {
 }
 
 void DeferredPipeline::render(const vector<uint> &cameras) {
-    if (cameras.empty()) return;
```
minggo, line 131: Why remove it?

cocos/bindings/dop/BufferAllocator.h
```diff
@@ -86,9 +84,10 @@ class CC_DLL BufferAllocator final : public cc::Object {
     static cc::vector<BufferAllocator *> pools;
     static constexpr uint                BUFFER_MASK = ~(1 << 30);
 
-    cc::map<uint, Object *>  _buffers;
-    cc::map<uint, uint8_t *> _caches;
-    PoolType                 _type = PoolType::UNKNOWN;
+    cc::map<uint, Object *> _buffers;
+    cc::vector<uint8_t *>   _array;
```
minggo, line 88: Should use a more meaningful name, such as _cachedBufferDataPointers.

cocos/bindings/dop/BufferAllocator.h
```diff
@@ -86,9 +84,10 @@ class CC_DLL BufferAllocator final : public cc::Object {
...
+    cc::vector<uint> _sizes;
```
minggo, line 89: The same as above, may use _cachedBufferLengths.

cocos/bindings/dop/BufferAllocator.h
caryliu1999: > Should use a more meaningful name, such as _cachedBufferDataPointers.

ok

cocos/renderer/pipeline/deferred/DeferredPipeline.cpp
caryliu1999: > Why remove it?

ZeQiang made a judgment at the JS layer, it is not needed here. And not calling flushCommands may cause multi-thread blocking.

minggo (approved):

PR #3445, opened 2021-04-22
add resize event in Metal

hana-alice:
tried to remove mu::renderEntireScreen, but failed in the _Camera_ Scene test. because MTLClear clear whole screen but here we need a viewport clear, scissor has no influence on the clear operation in Metal. the closest way to do this is https://developer.apple.com/documentation/metal/mtlrendercommandencoder/rendering_to_multiple_viewports_in_a_draw_command?language=objc , using multiple vp rendering in a single drawcall, but the _Camera_ test case seperate this scene into 4 passes, which makes it hard to apply multi-viewport-tech in.

### DIFF
--- a/cocos/platform/mac/View.mm
+++ b/cocos/platform/mac/View.mm
@@ -24,23 +24,26 @@
 ****************************************************************************/
 
 #import "View.h"
+#import <AppKit/NSEvent.h>
+#import <AppKit/NSScreen.h>
+#import <AppKit/NSTouch.h>
+#import <QuartzCore/QuartzCore.h>
 #import "KeyCodeHelper.h"
 #import "cocos/bindings/event/EventDispatcher.h"
 #include "platform/Application.h"
-#import <AppKit/NSEvent.h>
-#import <AppKit/NSTouch.h>
-#import <AppKit/NSScreen.h>
-#import <QuartzCore/QuartzCore.h>
 
 @implementation View {
-    cc::MouseEvent _mouseEvent;
+    cc::MouseEvent    _mouseEvent;
     cc::KeyboardEvent _keyboardEvent;
 }
 
 #ifdef CC_USE_METAL
-- (CALayer *)makeBackingLayer
-{
-    return [CAMetalLayer layer];
+- (CALayer *)makeBackingLayer {
+    CAMetalLayer *layer              = [CAMetalLayer layer];
+    layer.delegate                   = self;
+    layer.autoresizingMask           = true;
+    layer.needsDisplayOnBoundsChange = true;
+    return layer;
 }
 #endif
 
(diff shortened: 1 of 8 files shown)


cocos/renderer/gfx-metal/MTLCommandBuffer.mm
```diff
@@ -166,8 +167,15 @@ of this software and associated engine source code (the "Software"), a limited,
     else {
         _renderEncoder.initialize(_mtlCommandBuffer, mtlRenderPassDescriptor);
     }
-    _renderEncoder.setViewport(renderArea);
-    _renderEncoder.setScissor(renderArea);
+
+    Rect       scissorArea      = renderArea;
+#if defined(CC_DEBUG) && (CC_DEBUG > 0)
```
hana-alice, line 172: only debug mode there is a validation layer, renderArea is not consistent with rendertarget.w/h sometimes.
nativeWindow::resize->cc::EventDispatcher::resize->(...jsb call...)->js set camera width->(...jsb call...)->native beginrenderpass get renderArea. this long process makes a time delay, but rendertarget comes from drawable.texture directly, it's real time, thus inconsistent.

minggo: What will happen if they are consistent in release mode?

hana-alice: API validation is ignored, so no scissor area validation.

cocos/platform/mac/View.mm
```diff
@@ -49,15 +52,18 @@ - (instancetype)initWithFrame:(NSRect)frameRect {
         [self.window makeFirstResponder:self];
 
 #ifdef CC_USE_METAL
-        int pixelRatio = [[NSScreen mainScreen] backingScaleFactor];
-        CGSize size = CGSizeMake(frameRect.size.width * pixelRatio, frameRect.size.height * pixelRatio);
+        int    pixelRatio = [[NSScreen mainScreen] backingScaleFactor];
+        CGSize size       = CGSizeMake(frameRect.size.width * pixelRatio, frameRect.size.height * pixelRatio);
```
minggo, line 56: As frameRect includes pixelRatio in Application-mac.mm in this commit. So do not have to multiply it again.

cocos/renderer/gfx-metal/MTLCommandBuffer.mm
minggo: The effect is right?

hana-alice: yes, just ignore some assert, no other influence

cocos/platform/mac/View.mm
hana-alice: pixelRatio in appdeligate.mm reverted, to keep logic size consistent with other platform.

minggo (approved):

PR #3451, opened 2021-04-22
[3.1] add switch for NO_WERROR

PatriceJiang:
fix: https://github.com/cocos-creator/3d-tasks/issues/6979
fix: https://github.com/cocos-creator/3d-tasks/issues/6924

--- 

Set environment variable `COCOS_ENGINE_DEV` to `1`  to enable treat warning as error compilation flags.


PatriceJiang:
@minggo  麻烦合并下

PR #3452, opened 2021-04-22
fix iphone7p pixel ratio

hana-alice:
https://github.com/cocos-creator/3d-tasks/issues/6913

iphone7p special case:
https://www.zhihu.com/question/25288571


hana-alice:
paired with this one https://github.com/cocos-creator/engine/pull/8524

PR #3477, opened 2021-04-27
fix mtl pixelformat of pipelinestate when clear rect 

hana-alice:


cocos/renderer/gfx-metal/MTLUtils.mm
```diff
@@ -1626,5 +1630,16 @@ void main() {
     renderPassDescriptor.colorAttachments[slot].loadAction = MTLLoadActionLoad;
 }
 
+void mu::clearUtilResource() {
+    if(!renderPassMap.empty()) {
+        for (auto pass : renderPassMap) {
+            //TODO: create and destroy not in the same level
+            pass.second->destroy();
+            delete pass.second;
```
minggo, line 1638: Please use CC_DELETE as we may override delete and use our own memory management mechanism.

cocos/renderer/gfx-metal/MTLUtils.mm
```diff
@@ -39,6 +39,8 @@ of this software and associated engine source code (the "Software"), a limited,
 namespace gfx {
 namespace {
 
+std::unordered_map<uint, PipelineState*> renderPassMap;
```
minggo, line 42: It is better to define it in anonymous namespace.

minggo (approved):

PR #3478, opened 2021-04-27
fix application lifecycle issue on Android platform

PPpro:
fix: https://github.com/cocos-creator/3d-tasks/issues/6937

## ChangeLog:
- fix application lifecycle issue on Android platform

## Issue Description
- the js engine doesn't care about when the CocosAcitivity is paused
- the js engien care when the Android Application is paused (But the android API doesn't provide any app paused event listener)
- resuming the EditBoxActivity pauses the CocosActivity
- so when the `onPause` callback invoked in js engine, we don't know if it's because of the EditBoxActivity or because the app gets in the background

## How to Fix
reference: https://developer.android.com/guide/components/activities/activity-lifecycle
I think the `onStart` `onStop` lifecycle is exactly what we need in js engine  
the EditBoxActivity pauses the CocosActivity, but not stops it.
the `onStop` lifecycle is invoked only when the app gets in the background  

So we say that when the CocosActivity is stopped, the app's `onPause` callback should be invoked  
the same as the `onStart` phase

### DIFF
--- a/cocos/platform/android/jni/JniCocosActivity.cpp
+++ b/cocos/platform/android/jni/JniCocosActivity.cpp
@@ -68,9 +68,9 @@ int readCommand(int8_t &cmd) { //NOLINT(google-runtime-references)
 }
 
 void handlePauseResume(int8_t cmd) {
-    LOGV("activityState=%d", cmd);
+    LOGV("appState=%d", cmd);
     std::unique_lock<std::mutex> lk(cc::cocosApp.mutex);
-    cc::cocosApp.activityState = static_cast<uint8_t>(cmd);
+    cc::cocosApp.appState = static_cast<uint8_t>(cmd);
     lk.unlock();
     cc::cocosApp.cond.notify_all();
 }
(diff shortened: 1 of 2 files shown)


PPpro:
@minggo pls review

PPpro:
pending to merge into branch v3.2

minggo:
please fix coding style issue

cocos/platform/android/jni/JniCocosActivity.cpp
```diff
@@ -216,17 +216,17 @@ JNIEXPORT void JNICALL Java_com_cocos_lib_CocosActivity_onSurfaceCreatedNative(J
     setWindow(ANativeWindow_fromSurface(env, surface));
 }
 JNIEXPORT void JNICALL Java_com_cocos_lib_CocosActivity_onStartNative(JNIEnv *env, jobject obj) {
```
minggo, line 218: Function name should be named liked this to be invoked by java codes. So do not have to fix it, but should fix other coding style issues.

minggo (approved):

minggo:
@PatriceJiang i am not sure if it conflicts with your modification about CPU usage after entering background.

PatriceJiang (approved):

PR #3485, opened 2021-04-28
ios compiling error & mac full screen

hana-alice:
https://github.com/cocos-creator/3d-tasks/issues/7212

https://github.com/cocos-creator/3d-tasks/issues/7242


minggo (approved):

minggo:
Which modification will fix iOS compiling error?

hana-alice:
> Which modification will fix iOS compiling error?

"platform/device.h" -> upper case "platform/Device.h"
on some macbook, maybe a result of old-version xcode.

minggo:
Yep, and i think it is fixed in another PR.

minggo:
Oh, you closed that PR. I think it is better to fix iOS compiling error in a separate PR as it is more important than mac full screen issue.

PR #3504, opened 2021-05-07
V2.4.6 video

zxx43:
Add grab apis


minggo:
不是已经有 https://github.com/cocos-creator/engine-native/pull/3467 ？

PR #3529, opened 2021-05-18
dirty shadow map when size changed for v3.3

troublemaker52025:


troublemaker52025:
https://github.com/cocos-creator/engine/pull/8687

caryliu1999 (approved):

PR #3540, opened 2021-05-21
support SSPR by hw

dddpppccc:
1. support SSPR in cocos-native by HUAWEI
2. fix resize bug in deferredPipline
3. init depthAttachment rightly when subpass info is empty


minggo:
https://github.com/cocos-creator/engine-native/pull/3601

PR #3619, opened 2021-06-15
gfx: sRGB texture support

YunHsiao:
Also removed depth stencil format queries in gfx::Feature, all usages should directly pass `device->getDepthStencilFormat()` instead.

https://github.com/cocos-creator/3d-tasks/issues/7834


minggo:
Will it break API?

YunHsiao:
Yes the depth stencil format features are removed but as far as I know no one is using them.

PR #3626, opened 2021-06-17
[3.3] some little fixes

PatriceJiang:
- DummyGraphNode avoid bad link 
- Set Android default arch to arm64-v8a
- Android: do not copy asset during the build
- Speedup Travis CI Android build


PatriceJiang:
@minggo 麻烦合并下

PR #3638, opened 2021-06-22
[Add] Native scene object update function

GengineJS:


GengineJS:
https://github.com/cocos-creator/engine/pull/8863

PR #3669, opened 2021-06-29
crash prevention on Android 11 + API 30

xianyinchen:


minggo:
在 https://github.com/cocos-creator/engine-native/pull/3659 修改过了。

PR #3697, opened 2021-07-07
[Fix] Crash caused by real-time mesh generation on the native side

GengineJS:


GengineJS:
ts code：https://github.com/cocos-creator/engine/pull/8942

minggo (approved):

PatriceJiang (approved):

caryliu1999 (approved):

PR #3733, opened 2021-07-16
fix audio crash on iOS

PPpro:
Re: 
https://github.com/cocos-creator/3d-tasks/issues/6355
https://github.com/cocos2d/cocos2d-x/issues/18597

同步修复 pr： https://github.com/cocos2d/cocos2d-x/pull/18865

changeLog:
- 修复 iOS 端上频繁 play/stop 会 crash 的问题

### DIFF
--- a/cocos/audio/apple/AudioEngine-inl.mm
+++ b/cocos/audio/apple/AudioEngine-inl.mm
@@ -386,7 +386,7 @@ int AudioEngineImpl::play2d(const std::string &filePath, bool loop, float volume
 }
 
 void AudioEngineImpl::play2dImpl(AudioCache *cache, int audioID) {
-    //Note: It may bn in sub thread or main thread :(
+    //Note: It may be in sub thread or main thread :(
     if (!*cache->_isDestroyed && cache->_state == AudioCache::State::READY) {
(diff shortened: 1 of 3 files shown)


stevetranby:
Based on someone still seeing issues here: 
https://github.com/cocos2d/cocos2d-x/issues/18597#issuecomment-720988782

I'm curious if there needs to be a check inside the while loop for whether the source is still playing or looping? If destroy is called, I presume a looping sound should also end?

It seems like maybe the `while(buffers < MAX)` may cause infinite loop? I'm guessing OpenAL almost always executes in a manner where it will eventually process more than MAX buffers ending the while loop, but if `alGetSourcei(_alSource, AL_BUFFERS_PROCESSED, &bufferProcessed);` always returns a value of 0 after it reaches that iOS specific section then it does seem like it will deadlock at that point.

Could it be the case that you should only process up to MAX buffers (using `while(bufferProcessed > 0) { bufferProcessed--; } ` like elsewhere in the audio engine) ??

I'm going to wait to see if I get any more crashes in app store versions, but wanted to just see if you had any thoughts on the changes in this commit.

PR #3738, opened 2021-07-16
Fix urlopen error in download-deps.py caused by missing TSL 1.2 support

JasonHuang3D:
- **TSL 1.2** is only available for python **2.7.9 or later**, so when using the earlier version and trying to connect the severs (such as github)  which are using TSL 1.2 by using **urlopen** will lead an error:
error

- This fix added the python full version check after major version of 2.x validation.

### DIFF
--- a/download-deps.py
+++ b/download-deps.py
@@ -304,6 +304,12 @@ def _check_python_version():
                "Download it here: https://www.python.org/" % (major_ver, sys.version_info[1]))
         return False
 
+    # To make sure TLS 1.2 is supported by python
+    if sys.version_info < (2, 7, 9):
+        print ("The python version is %d.%d.%d. Please install 2.7.9 or higher version of python 2.x\n"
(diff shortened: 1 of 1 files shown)


download-deps.py
```diff
@@ -304,6 +304,12 @@ def _check_python_version():
                "Download it here: https://www.python.org/" % (major_ver, sys.version_info[1]))
         return False
 
+    # To make sure TLS 1.2 is supported by python
+    if sys.version_info < (2, 7, 9):
+        print ("The python version is %d.%d.%d. Please install 2.7.9 or higher version of python 2.x\n"
+               "Download it here: https://www.python.org/" % (sys.version_info[0], sys.version_info[1], sys.version_info[2]))
+        return False
+
```
JasonHuang3D, line 312: Let me know if the print message is good for you guys.

pandamicro (approved):

PR #3739, opened 2021-07-17
Fix se::Value dose not accept unsigned int 64.

JasonHuang3D:
se::Value dose not accept unsigned int 64,  so cast to unsigned int32 for 64 bit build to solve a build error in 64 bit build.

### DIFF
--- a/cocos/bindings/manual/jsb_cocos_manual.cpp
+++ b/cocos/bindings/manual/jsb_cocos_manual.cpp
@@ -642,7 +642,7 @@ static bool js_engine_FileUtils_listFilesRecursively(se::State &s) {
         for (uint i = 0; i < static_cast<uint>(arg1.size()); i++) {
             list->setArrayElement(i, se::Value(arg1[i]));
         }
-        list->setProperty("length", se::Value(arg1.size()));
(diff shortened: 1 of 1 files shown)


cocos/bindings/manual/jsb_cocos_manual.cpp
```diff
@@ -642,7 +642,7 @@ static bool js_engine_FileUtils_listFilesRecursively(se::State &s) {
         for (uint i = 0; i < static_cast<uint>(arg1.size()); i++) {
             list->setArrayElement(i, se::Value(arg1[i]));
         }
-        list->setProperty("length", se::Value(arg1.size()));
+        list->setProperty("length", se::Value(static_cast<std::uint32_t>(arg1.size())));
```
JasonHuang3D, line 645: static_cast will be optimized by compiler if builds to 32 bit, so there is no overhead for 32 bit build.

PatriceJiang:
Please commit to branch 3.3.0

JasonHuang3D:
@PatriceJiang Hi, the branch 3.3.0 seems to be very different than develop branch. My issue can not be resolved in that branch. I believe this PR only make sense to develop branch.

PatriceJiang (approved):

PR #3740, opened 2021-07-17
Fix SetWindowLongPtr for native windows.

JasonHuang3D:
- Fix a build error which is caused by passing wrong argument to Windows API.

- As per [MSDN](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowlongptrw) document, the second argument of SetWindowLongPtr function dose not accept **GWL_WNDPROC** but **GWLP_WNDPROC**.  

- Generally, for Windows API   "LP" stands for long pointer,  "L" is stands for long (integer value). Thus:
  - SetWindowLong uses GWL_WNDPROC.
  - SetWindowLongPtr  uses GWLP_WNDPROC.

### DIFF
--- a/cocos/ui/edit-box/EditBox-win32.cpp
+++ b/cocos/ui/edit-box/EditBox-win32.cpp
@@ -183,8 +183,8 @@ void EditBox::show(const EditBox::ShowInfo &showInfo) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


PatriceJiang:
Please commit to branch 3.3.0

JasonHuang3D:
@PatriceJiang  Hi,  the branch 3.3.0 seems to be very different than develop branch. My issue can not be resolved in that branch. I believe this PR only make sense to develop branch.

PatriceJiang (approved):

PR #3752, opened 2021-07-21
mem leaks

hana-alice:
metal mem leaks


minggo:
关联一下 issue 吧

hana-alice:
https://github.com/cocos-creator/3d-tasks/issues/7818

PR #3766, opened 2021-07-26
fix command buffer agent

YunHsiao:
`CommandBufferAgent.updateBuffer` should only invoke itself on actors, not cross invoking `BufferAgent.update` instead.

This will cause buffers to be unconditionally immediately updated, which is not what we want.

### DIFF
--- a/cocos/renderer/gfx-agent/BufferAgent.h
+++ b/cocos/renderer/gfx-agent/BufferAgent.h
@@ -42,7 +42,7 @@ public:
 
     void update(const void *buffer, uint size) override;
 
-    void update(const void *buffer, uint size, MessageQueue *mq);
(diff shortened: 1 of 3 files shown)


cocos/renderer/gfx-agent/BufferAgent.h
```diff
@@ -42,7 +42,7 @@ class CC_DLL BufferAgent final : public Agent<Buffer> {
 
     void update(const void *buffer, uint size) override;
 
-    void update(const void *buffer, uint size, MessageQueue *mq);
+    static void getActorBuffer(const BufferAgent *buffer, MessageQueue *mq, uint size, uint8_t **pActorBuffer, bool *pNeedFreeing);
```
minggo, line 45: pNeedFreeing：目前引擎是没有用这种匈牙利命名法的。

minggo (approved):

hana-alice (approved):

PR #3791, opened 2021-08-02
Fix the incorrect ModelBounds of BakedSkinningModel

GengineJS:


GengineJS:
ts: https://github.com/cocos-creator/engine/pull/9113

minggo (approved):

PR #3792, opened 2021-08-02
Update cmake list for dynamic library update

timlyeee:
* as glslang and tbb become dynamic library, the cmake should copy dll from glslang folder too.
* "gulp gen simulator" command will automatically generate by user's comupter so sometime it will generate x64 version's simulator. no it's update to  generate Win32 version
* Edit-Box was using wrong parameters GWL_WNDPROC, so x64 version will hit an error. details below:https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowlongptra

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2073,6 +2073,7 @@ if(OHOS)
         rawfile.z
         z
         OpenSLES
+        hilog_ndk.z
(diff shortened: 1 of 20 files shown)


timlyeee:
目前external还没审过，tag没有无法下载

PR #3815, opened 2021-08-10
correct lighting stage attachment configuration

hana-alice:
https://github.com/cocos-creator/3d-tasks/issues/8547#event-5128822486


hana-alice:
already fixed

PR #3821, opened 2021-08-12
V3.3.0 memory leak

stanleyljl:
Add memory leak detector system for Cocos Creator: https://github.com/cocos-creator/3d-tasks/issues/8575


minggo:
这个 PR 还有用吗？

PR #3865, opened 2021-09-03
fix fg exec storage

YunHsiao:
The capture groups are lost whenever the execute lambda is passed as lvalue references.

### DIFF
--- a/cocos/renderer/gfx-base/GFXSwapchain.cpp
+++ b/cocos/renderer/gfx-base/GFXSwapchain.cpp
@@ -52,12 +52,18 @@ void Swapchain::destroy() {
 }
 
 void Swapchain::resize(uint32_t width, uint32_t height, SurfaceTransform transform) {
-    if (width != _colorTexture->getWidth() || height != _colorTexture->getHeight()) {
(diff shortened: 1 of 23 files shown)


cocos/renderer/gfx-base/GFXSwapchain.cpp
```diff
@@ -52,12 +52,18 @@ void Swapchain::destroy() {
...
-        if (isPreRotationEnabled()) _transform = transform;
+        // if (isPreRotationEnabled()) {
+        //     if (toNumber(transform) % 2) {
+        //         std::swap(width, height);
+        //     }
+        //     _transform = transform;
+        // }
+
+        // _colorTexture->_info.width = _depthStencilTexture->_info.width = width;
+        // _colorTexture->_info.height = _depthStencilTexture->_info.height = height;
```
minggo, line 66: 调试代码？

minggo (approved):

cocos/renderer/gfx-base/GFXSwapchain.cpp
YunHsiao: 在李欣那的 resize 事件重构前，需要暂时注掉

PR #3886, opened 2021-09-14
Fix swapchain resize issue.

timlyeee:
* If the new surface created by resize event is using the same address with the one store  in context, error occurs
* Should set to EGL_NO_SURFACE when destroy the original surface


YunHsiao (approved):

minggo (approved):

timlyeee:
https://github.com/cocos-creator/3d-tasks/issues/8213

PR #3888, opened 2021-09-15
should trigger update when set instance

YunHsiao:
https://github.com/cocos-creator/3d-tasks/issues/9156
https://github.com/cocos-creator/3d-tasks/issues/9139


pandamicro:
Please also confirm https://github.com/cocos-creator/3d-tasks/issues/9139

PR #3913, opened 2021-09-24
[3.4] Android fix crash on startup

PatriceJiang:
revert: https://github.com/cocos-creator/engine-native/pull/3886

Update cache of EGL read/write surface.


PatriceJiang:
@timlyeee 麻烦验证下能否解决之前的问题

PatriceJiang:
这个 pr 不解决问题，关闭

PR #3929, opened 2021-09-28
Update config.json

timlyeee:
https://github.com/cocos-creator/engine-native-external/pull/193


timlyeee:
for windows : 
- ignore static_cast warning as error for v8-internal.h : it will cast unsigned long long to unsigned int in x86 version

PR #3954, opened 2021-10-08
Fix close window crash for v3.4.0

troublemaker52025:


troublemaker52025:
* https://github.com/cocos-creator/3d-tasks/issues/9263

PR #4008, opened 2021-10-26
Fix restart issue 

timlyeee:
restart issue contains 2 principal backends.

## GLES:

After restart, the context was not destroyed totally, but the swapchain was recreated by using the old context of Device, while the data is out of date.

It's good to re-init the context but only gl backend has this bug. Temperature solution is to destroy the device. It will be recreated after then.

## VK

The surface should be destroyed once we put it background, but at windows platform, the default custom event did not offer the WindowHandler. 

So hack it.


timlyeee:
需要engine的修改才能完整的运行该改动。插屏界面在3.5会做大的调整，所以目前engine只做小的hack处理。
* https://github.com/cocos-creator/engine/pull/9606

此外，和框架组确认在最小化时主循环其实也是暂停的，也就是说本身就不支持挂机行为

timlyeee:
该部分修改将在3.5统一接口后被重构
* https://github.com/cocos-creator/3d-tasks/issues/10031

minggo (approved):

YunHsiao (approved):

PR #4023, opened 2021-11-01
Solve the problem of batched buffer memory leak

GengineJS:


minggo (approved):

cocos/renderer/pipeline/BatchedBuffer.cpp
```diff
@@ -42,6 +42,19 @@ BatchedBuffer *BatchedBuffer::get(scene::Pass *pass, uint extraKey) {
     return buffer;
 }
 
+void BatchedBuffer::destroyBatchedBuffer() {
+    for (auto &pair : buffers) {
```
minggo, line 46: buffers -> BatchedBuffer:: buffers 会更清晰点

GengineJS: 修改了

star-e (approved):

PR #4027, opened 2021-11-02
fix iphone 5s compile issue

hana-alice:


hana-alice:
https://github.com/cocos-creator/3d-tasks/issues/9573

PR #4029, opened 2021-11-02
Add ENABLE_SANDBOX option

timlyeee:
Add option for compile step


timlyeee:
The reason why mac will crash for some reason is not found, to make it controllable I add an option to add sandbox manually.

PR #4073, opened 2021-11-30
Remove task detect code, use singleTask only for android4.4+

timlyeee:
## Modification
* No need to use task detect code anymore, `singleTask` property can fit it. 
* Export game activity, so that the game can be opened outside. It is also forces required in Android API 31 (Android 13) 

## Why?

If we don't remove `taskAffinity` property, the developer will get confused when he tried to use another activity to open the game activity, and it will open 2 tasks at the same time, but the developer might not figure out why. And it also makes SDK like Admob hard to integrate.

The property `android:exported: "true"` is required in Android 13, which is released in Nov.2021, and it also makes one situation possible: Open the game from Taptap or Google play.

### DIFF
--- a/templates/android/template/app/AndroidManifest.xml
+++ b/templates/android/template/app/AndroidManifest.xml
@@ -24,7 +24,7 @@
             android:label="@string/app_name"
             android:theme="@android:style/Theme.NoTitleBar.Fullscreen"
             android:launchMode="singleTask"
-            android:taskAffinity="" >
+            android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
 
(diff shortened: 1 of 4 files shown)


minggo (approved):

minggo:
> No need to use task detect code anymore, singleTask property can fit it.

Why need to set it before? Is it because it only has problem in older Android OS versions?

timlyeee:
> > No need to use task detect code anymore, `singleTask` property can fit it.
> 
> Why need to set it before? Is it because it only has problems in older Android OS versions?

Yes, codes to detect `isTaskRoot` prevent if the installer or other app opens the game activity, and the user still tries to open it again, which means the task should always be **`single`** and **`root`**. However, in **newer OS** the property `singleTask` can make the task **`single`**, the game activity is required to be exported so it should not be root either. 

Thus codes have no reason to exist.

PR #4111, opened 2021-12-18
Correct visibility type

pandamicro:
https://github.com/cocos-creator/3d-tasks/issues/10518


pandamicro:
Fix issue caused by https://github.com/cocos-creator/engine-native/pull/4098

PR #4170, opened 2022-01-13
fix iOS keyboard hide issue.

xianyinchen:
iOS keyboard hiding may be in the word selection screen. so check the offset of keyboard, if keyboard no work, offset <= 0

**BUG, Scrolling to select a word will exit the keyboard directly.**
image


minggo:
Would you please sync to v3.x?

timlyeee (approved):

minggo:
@xianyinchen 帮忙同步到 3.x 。

PR #4172, opened 2022-01-13
Revert "allow tick() in background (#4083)"

timlyeee:
This reverts commit 1191c1cf4e88af80dcc31a44ccb952b2e07362d4.

which will cause fault display when app create prefeb in background

* https://github.com/cocos-creator/3d-tasks/issues/10768


minggo:
已经合并了 https://github.com/cocos-creator/engine-native/pull/4193，所以关闭该 PR

PR #4184, opened 2022-01-17
Optimising the initialisation timing of download tasks.

xianyinchen:
alway check null value before use it, and this will fix some issue like no response return to business code when downloading a file.


PatriceJiang (approved):

PatriceJiang:
麻烦合并到 3.x

PR #4169, opened 2022-01-13
add a new interface to query format feature info

yiwenxue:
Re: cocos-creator/3d-tasks#8203 

TS engine: https://github.com/cocos-creator/engine/pull/9949#issue-1096192951

Move the query of format feature information to the new interface. In the old method, one can only know whether the device generally supports a feature, but the fact is that some formats support this feature, and others don't. With the new interface, one can query whether a feature is supported by a specific format or not.

Vulkan and GLES are tested with test-cases-3d. This modification introduced no extra errors.

PR #4223, opened 2022-01-25
V3.6 merge v3.5.0

minggo:


YunHsiao:
@minggo #4169 was merged yesterday, need to sync again

minggo:
Does engine includes it too? If not, then i will sync it in another PR as this PR if very complex now, and have many conflicts.

YunHsiao:
Yes, both repos need to be synced

minggo:
I think we can sync it later.

minggo:
合并分支不处理 clang-tidy 问题。

PR #4230, opened 2022-01-27
more validations

YunHsiao:
Rely on: https://github.com/cocos-creator/engine/pull/10058

Added more validations on error-prone details in gfx public interfaces


minggo:
I think it is better to send to v3.6.

PR #4240, opened 2022-02-08
Fix android and ohos load library bug

timlyeee:
* Should keep same name as cmake 
* set param to ${ENGINE_NAME}, easier to modify


PatriceJiang (approved):

templates/android/template/CMakeLists.txt
```diff
@@ -17,8 +17,8 @@ set(PROJ_SOURCES
 add_library(${LIB_NAME} SHARED ${PROJ_SOURCES})
 
 target_link_libraries(${LIB_NAME}
-    "-Wl,--whole-archive" cocos_jni "-Wl,--no-whole-archive"
-    cocos
+    "-Wl,--whole-archive" cocos2d_jni "-Wl,--no-whole-archive"
```
minggo, line 20: 这里为什么还是 cocos2d 呢？

minggo (approved):

PR #4254, opened 2022-02-10
Fix loop audio would stop for ios system 

timlyeee:
* IOS system has an uncertain state between foreground and background, and our condition sometimes miss the point to load buffers.
* Queue for 3 buffers might be too short to live, use 4 or even more instead.
* Issue could be solved: 
* https://github.com/cocos2d/cocos2d-x/issues/19480, 
* https://github.com/cocos-creator/3d-tasks/issues/11135

### DIFF
--- a/cocos/audio/apple/AudioPlayer.mm
+++ b/cocos/audio/apple/AudioPlayer.mm
@@ -244,7 +244,12 @@ void AudioPlayer::rotateBufferThread(int offsetFrame) {
 
         while (!_isDestroyed) {
             alGetSourcei(_alSource, AL_SOURCE_STATE, &sourceState);
-            if (sourceState == AL_PLAYING) {
+            /* On IOS, audio state will lie, when the system is not fully foreground,
+             * openAl will process the buffer in queue, but our condition cannot make sure that the audio
+             * is playing as it's too short. Interesting IOS system.
+             * Solution is to load buffer even if it's paused, just make sure that there's no bufferProcessed in 
+             */
+            if (sourceState == AL_PLAYING || sourceState == AL_PAUSED) {
                 alGetSourcei(_alSource, AL_BUFFERS_PROCESSED, &bufferProcessed);
                 while (bufferProcessed > 0) {
                     bufferProcessed--;
--- a/cocos/audio/apple/AudioMacros.h
+++ b/cocos/audio/apple/AudioMacros.h
@@ -26,7 +26,7 @@
 
 #pragma once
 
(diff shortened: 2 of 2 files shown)


cocos/audio/apple/AudioPlayer.mm
```diff
@@ -244,7 +244,12 @@ of this software and associated engine source code (the "Software"), a limited,
 
         while (!_isDestroyed) {
             alGetSourcei(_alSource, AL_SOURCE_STATE, &sourceState);
-            if (sourceState == AL_PLAYING) {
+            /* On IOS, audio state will lie, when the system is not fully foreground,
+             * openAl will process the buffer in queue, but our condition cannot make sure that the audio
+             * is playing as it's too short. Interesting IOS system.
+             * Solution is to load buffer even if it's paused, just make sure that there's no bufferProcessed in 
+             */
+            if (sourceState == AL_PLAYING || sourceState == AL_PAUSED) {
```
minggo, line 252: Load buffer even it is paused? Will it exceed the buffer?

timlyeee: No, only processed buffers will be reloaded.

minggo (approved):

dumganhar:
Ref：https://github.com/cocos-creator/engine/pull/10094

dumganhar (approved):

frozax:
It didn't fix the issue in my case (looping audio stops randomly, not necessarily at the loop section).

timlyeee:
> It didn't fix the issue in my case (looping audio stops randomly, not necessarily at the loop section).

@frozax Any issue related? Or could you describe it with details?

frozax:
> > It didn't fix the issue in my case (looping audio stops randomly, not necessarily at the loop section).
> 
> @frozax Any issue related? Or could you describe it with details?

You're right it's missing some information, I'm sorry. On my 9 games, only one has music. It's a loop of 49 seconds. On iOS, it's a wav file of 4.19 Mb. I'm using cocos2dx 3.17.2, `experimental::AudioEngine`. On Android, I don't have any issue (using a 356k ogg of 49 seconds).
I've seen it happen regularly on a iPad Mini running iOS 11 (I think, I'll have to check with the owner), it happens usually between 1-10 minutes of gameplay. Once the music stopped, the `AudioEngine::getState` return that it's NOT `AudioState::PLAYING` (I can't tell more) and I can restart it again.
Hope this helps

timlyeee:
Thank you for your information, we will try it.

PR #4255, opened 2022-02-11
select mipmap levels dynamically

yiwenxue:
Re: cocos-creator/3d-tasks#7952

Add TextureView support to all GFX backends, so that one can access Image through texture, which specifies the approach and the mipmap-level range. A TextureView contains a pointer to Texture, `type`, `format`, `baseLevel`, and `levelCount`. The render will only use texture mipmaps ranging from `baseLevel` to `baseLevel + levelCount`.


minggo:
3.5 要提交到 engine 仓库了。

PR #4264, opened 2022-02-23
[2.4] Upgrade okhttp 3.12.7 -> 3.12.14

PatriceJiang:
Download link https://github.com/PatriceJiang/okhttp/blob/cocos2dx-rename-3.12.x/libs/okhttp-3.12.14-SNAPSHOT.jar

Re: https://forum.cocos.org/t/topic/119244/7


minggo:
是否得同步到 3.x？

PatriceJiang:
已同步到 https://github.com/cocos-creator/engine/pull/10169

PR #4271, opened 2022-03-02
fix spine crash

xianyinchen:
without call Object::clearPrivateData, the c++ object maybe bind to two JS object, when one was release, other one will  case crash.

Some existent crash stacks.

企业微信截图_16461276994724
企业微信截图_16461323241416
企业微信截图_16461880675214


minggo:
3.4.2 来不及了，后续得提交到 3.5。

PatriceJiang:
@xianyinchen dragonbone 也有同样的问题

PR #4279, opened 2022-03-17
Avoid mask case crash on some phone. 

pandamicro:


pandamicro:
Sync https://github.com/cocos/engine-native/pull/3698

PR #4290, opened 2022-07-19
get language by using system language settings

timlyeee:
re: https://github.com/cocos/cocos-engine/issues/10107

Call to get language by calling LocaleList after Android7, as it add region check and bugs occurs

The bug does not show before Android7


cocos/platform/android/java/src/org/cocos2dx/lib/Cocos2dxHelper.java
```diff
@@ -302,13 +302,21 @@ public static String getPackageName() {
     public static String getWritablePath() {
         return Cocos2dxHelper.sFileDirectory;
     }
-
+
```
minggo, line 305: extra spaces

PR #4311, opened 2023-04-15
fix blend state set incorrect when blendSeparation

yanjifa:


minggo:
@yanjifa thanks for your contribution.
@hana-alice could you please take a look.

hana-alice:
LGTM, @bluesky013 any suggestions？

bluesky013 (approved):

hana-alice (approved):

PR #4317, opened 2023-09-05
Openharmony build templates

qiuguohua:


templates/js-template-default/frameworks/runtime-src/proj.openharmony/CMakeLists.txt
```diff
@@ -12,8 +12,10 @@ project(${APP_NAME} CXX)
 
 set(CMAKE_CXX_FLAGS_DEBUG  "-Werror=return-type -ferror-limit=0")
 
+enable_language(C ASM)
+set(CMAKE_CXX_STANDARD 17)
```
minggo, line 16: v2.x also need c++17?

qiuguohua: std::option is used in the napi.

minggo: Ok, as OH support c++17, there is not problem.

minggo (approved):

PR #4347, opened 2024-05-17
fixed #2d-tasks/issues/3854: iOS LaunchScreen.storyboard compilation error

dumganhar:
https://github.com/cocos/2d-tasks/issues/3854

### DIFF
--- a/templates/js-template-default/frameworks/runtime-src/proj.ios_mac/ios/Base.lproj/LaunchScreen.storyboard
+++ b/templates/js-template-default/frameworks/runtime-src/proj.ios_mac/ios/Base.lproj/LaunchScreen.storyboard
@@ -4,7 +4,7 @@
         <adaptation id="fullscreen"/>
     </device>
     <dependencies>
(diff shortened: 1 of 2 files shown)


templates/js-template-default/frameworks/runtime-src/proj.ios_mac/ios/Base.lproj/LaunchScreen.storyboard
```diff
@@ -4,7 +4,7 @@
         <adaptation id="fullscreen"/>
     </device>
     <dependencies>
-        <deployment version="1792" identifier="iOS"/>
+        <deployment version="2048" identifier="iOS"/>
```
dumganhar, line 7: 2048 means iOS 8 or later

dumganhar: WeChat79d2ecd79c586f1c3be19cd6f7f0e605

minggo (approved):

PR #4360, opened 2024-07-22
fix iterator maybe invalid.

benloong:
spine object dispose callback could recursively finalize , leading iterator invalid.


minggo (approved):

minggo:
@benloong could you please send to v2.4.15? Thanks.

benloong:
@minggo sure.

minggo:
thanks

PR #4368, opened 2024-10-17
Change OpenRawFile to 64 interface and fix some bugs

wanghui187:
Change OpenRawFile to 64 interface
Remove debug compilation options
Add jsvm.so compilation link judgment
Remove Utels.cpp and compile reference error in V8


cocos/audio/android/AudioEngine-inl.cpp
```diff
@@ -132,7 +132,7 @@ static int fdGetter(const std::string& url, off_t* start, off_t* length)
     #elif CC_TARGET_PLATFORM == CC_PLATFORM_OPENHARMONY
     FileUtilsOpenHarmony* fileUtils = static_cast<FileUtilsOpenHarmony*>(FileUtils::getInstance());
     if(fileUtils) {
-        RawFileDescriptor descriptor;
+        RawFileDescriptor64 descriptor;
```
qiuguohua, line 135: Can this interface support 32-bit? We have the RK3568 development board for 3.8 that supports 32-bit.

PR #4375, opened 2024-11-28
Modify the issue of calling restart and hide twice

wanghui187:
Modify the issue of calling restart and hide twice


templates/js-template-default/frameworks/runtime-src/proj.harmonyos-next/entry/src/main/ets/entryability/EntryAbility.ts
```diff
@@ -7,6 +7,7 @@ const nativeContext = cocos.getContext(ContextType.ENGINE_UTILS);
 const nativeAppLifecycle = cocos.getContext(ContextType.APP_LIFECYCLE);
 
 export default class EntryAbility extends UIAbility {
+  private windowStageType: number = 0;
```
qiuguohua, line 10: Is it possible to use enums here?

PR #4376, opened 2024-11-28
Modify the issue of calling restart and hide twice

wanghui187:
Modify the issue of calling restart and hide twice

### DIFF
--- a/cocos/scripting/js-bindings/jswrapper/v8/ScriptEngine.cpp
+++ b/cocos/scripting/js-bindings/jswrapper/v8/ScriptEngine.cpp
@@ -140,7 +140,7 @@ namespace se {
             if ((argc - msgIndex) == 1)
             {
                 std::string msg = args[msgIndex].toStringForce();
-                SE_LOGD("JS: %s%s\n", prefix, msg.c_str());
+                SE_LOGD("JS: %{public}s%{public}s\n", prefix, msg.c_str());
             }
             else if (argc > 1)
             {
@@ -159,7 +159,7 @@ namespace se {
                     }
                 }
 
-                SE_LOGD("JS: %s%s\n", prefix, msg.c_str());
+               SE_LOGD("JS: %{public}s%{public}s\n", prefix, msg.c_str());
             }
 
             return true;
@@ -461,8 +461,8 @@ namespace se {
         if(!jitSupported()) {
             flags.append(" --jitless");
         }
-#elif CC_TARGET_PLATFORM == CC_PLATFORM_OPENHARMONY
-        flags.append(" --jitless");
+#elif (CC_TARGET_PLATFORM == CC_PLATFORM_OPENHARMONY)
+            flags.append(" --jitless");
 #endif
         if(!flags.empty())
         {
--- a/cocos/scripting/js-bindings/manual/jsb_websocket.cpp
+++ b/cocos/scripting/js-bindings/manual/jsb_websocket.cpp
@@ -272,6 +272,7 @@ static bool WebSocket_finalize(se::State& s)
         cobj->closeAsync();
     }
     if(se::ScriptEngine::getInstance()->isInCleanup()) {
+        cobj->release();
         return true;
     }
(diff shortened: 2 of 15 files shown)


qiuguohua:
https://github.com/cocos/2d-tasks/issues/3844

templates/js-template-default/frameworks/runtime-src/proj.harmonyos-next/entry/src/main/ets/entryability/EntryAbility.ts
```diff
@@ -5,8 +5,13 @@ import window from '@ohos.window';
 
 const nativeContext = cocos.getContext(ContextType.ENGINE_UTILS);
 const nativeAppLifecycle = cocos.getContext(ContextType.APP_LIFECYCLE);
-
+enum windowStageType {
+  hide,
+  show
+}
 export default class EntryAbility extends UIAbility {
+  private windowStageType: number = 0;
```
qiuguohua: Should this type be an enum?

wanghui187: fix

qiuguohua (approved):

cocos/scripting/js-bindings/manual/jsb_websocket.cpp
```diff
@@ -271,7 +271,9 @@ static bool WebSocket_finalize(se::State& s)
         CCLOGINFO("WebSocket (%p) isn't closed, try to close it!", cobj);
         cobj->closeAsync();
     }
-
+    if(se::ScriptEngine::getInstance()->isInCleanup()) {
+        return true;
+    }
```
minggo: Why add it? Will it cause memory leak?

cocos/scripting/js-bindings/jswrapper/v8/ScriptEngine.cpp
```diff
@@ -461,8 +461,8 @@ namespace se {
         if(!jitSupported()) {
             flags.append(" --jitless");
         }
-#elif CC_TARGET_PLATFORM == CC_PLATFORM_OPENHARMONY
-        flags.append(" --jitless");
+#elif (CC_TARGET_PLATFORM == CC_PLATFORM_OPENHARMONY)
```
minggo, line 464: Can add it in above if, like 
```
#if (CC_TARGET_PLATFORM == CC_PLATFORM_IOS || CC_TARGET_PLATFORM == CC_PLATFORM_OPENHARMONY)
```

cocos/scripting/js-bindings/manual/jsb_websocket.cpp
wanghui187: We found that there is an issue with the double free pointer when restarting in jsvm, and it can only be returned here. This modification does not affect other scenarios

cocos/scripting/js-bindings/jswrapper/v8/ScriptEngine.cpp
wanghui187: That's it, iOS still has jitSupported judgment

minggo: Ok

cocos/scripting/js-bindings/manual/jsb_websocket.cpp
minggo: But release is not invoked. I don't think it is a good way to resolve the issue for jsvm.

wanghui187: fix，I made some modifications and called release

minggo: The logic is different from original. You should find the reason why it is double released.

minggo (approved):

cocos/scripting/js-bindings/manual/jsb_websocket.cpp
wanghui187: Fixed

PR #4386, opened 2025-03-20
Fix 'Object::createTypedArrayWithBuffer' for JSVM.

dumganhar:
It's a bug starts from the first version of JSVM support in https://github.com/cocos/engine-native/pull/4366 .

Re: https://github.com/cocos/3d-tasks/issues/18837

https://github.com/cocos/cocos-engine/pull/18487

3.x napi backend was fixed at https://github.com/cocos/cocos-engine/pull/14743


qiuguohua:
@dumganhar Napi needs to be modified too.

qiuguohua (approved):

PR #4390, opened 2025-04-14
Fix the probabilistic crash of JSVM

lynnllh:
When the JSVM execution fails, the value of return_val remains the random value allocated on the stack during initialization. Then, when the OH_JSVM_Typeof method is executed, there is a probability of a crash occurring.


dumganhar:
@lynnllh Good catch.

Since v2.4.15 was released, I will change the target branch of this PR to v2.4.16

dumganhar:
@lynnllh Merged. Thanks for your contribution. I will cherry-pick this commit to Creator v3.

PR #4392, opened 2025-04-21
Fix audio problems in openharmony

lynnllh:
The OH_AudioRenderer_OnWriteData callback of ohaudio does not support determining whether playback is required based on the return value. On some system versions, such as 5.0.0.135 (SP5), there may be issues with abnormal audio playback. It is recommended to switch to the OH_AudioRenderer_OnWriteDataCallback function on api12. The relevant link is: https://developer.huawei.com/consumer/cn/doc/harmonyos-guides/using-ohaudio-for-playback

### DIFF
--- a/cocos/audio/openharmony/PcmAudioService.cpp
+++ b/cocos/audio/openharmony/PcmAudioService.cpp
@@ -47,7 +47,7 @@ PcmAudioService::~PcmAudioService() {
 }
 
 
-int32_t PcmAudioService::AudioRendererOnWriteData(OH_AudioRenderer* renderer,
+OH_AudioData_Callback_Result PcmAudioService::AudioRendererOnWriteData(OH_AudioRenderer* renderer,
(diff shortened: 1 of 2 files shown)


cocos/audio/openharmony/PcmAudioService.cpp
```diff
@@ -47,7 +47,7 @@ PcmAudioService::~PcmAudioService() {
 }
 
 
-int32_t PcmAudioService::AudioRendererOnWriteData(OH_AudioRenderer* renderer,
+OH_AudioData_Callback_Result PcmAudioService::AudioRendererOnWriteData(OH_AudioRenderer* renderer,
```
qiuguohua, line 50: I think we can change it to a global function later, so there's no need to reference it in the header file

qiuguohua (approved):

star-e (approved):

PR #4393, opened 2025-04-23
Fix the issue of incorrect calculation of the reserved area in openharmony

lynnllh:
Fix the issue of incorrect calculation of the reserved area on some tablets (PCE, where the notch screen is in the middle position)


templates/js-template-default/frameworks/runtime-src/proj.harmonyos-next/entry/src/main/ets/cocos/sys-ability-polyfill.ets
```diff
@@ -135,35 +145,74 @@ globalThis.initScreenInfo = () => {
...
+        return 0;
+    }
+
+    return cutOut.left + cutOut.width;
+}
+
+globalThis.getAvoidSide = () => {
+    if (!sceneInfo.width || !sceneInfo.height) {
+        console.log("get scene info error!");
+        return 0;
```
qiuguohua: Can we return -1 here because 0 is a reasonable value?

lynnllh: Returning 0 was intended to default to avoiding upward, but after thinking about it, defaulting to not avoiding might be a better choice.

lynnllh: > Can we return -1 here because 0 is a reasonable value?

It has been modified, please help review again.

star-e (approved):

cocos/platform/openharmony/CCDevice-openharmony.cpp
```diff
@@ -55,36 +55,47 @@ void Device::setKeepScreenOn(bool value) {
...
+        }
+        safearea_bottom += cutout_bottom;
+    } else if(3 == avoidSide) {
+        auto left = NapiHelper::napiCallFunction("getCutoutToLeft");
+        int32_t cutout_left = 0;
+        if (left.IsNumber()) {
+            cutout_left = left.As<Napi::Number>().Int32Value();
+        }
+        safearea_left += cutout_left;
+    }
```
qiuguohua, line 97: Should we handle the situation of returning -1 here?

templates/js-template-default/frameworks/runtime-src/proj.harmonyos-next/entry/src/main/ets/cocos/sys-ability-polyfill.ets
qiuguohua: Returning -1 can reduce one call to getCutoutToTop

cocos/platform/openharmony/CCDevice-openharmony.cpp
lynnllh: > Should we handle the situation of returning -1 here?
If the screen width and height are not obtained, -1 will be returned. At this point, it is unclear which direction to avoid. I think in this case, we can ignore the -1 situation, meaning no avoidance by default. What do you think?

lynnllh: > Should we handle the situation of returning -1 here?


If the screen width and height are not obtained, -1 will be returned. At this point, it is unclear which direction to avoid. I think in this case, we can ignore the -1 situation, meaning no avoidance by default. What do you think?

qiuguohua: I agree with your idea. Default is no avoidance and output error log

qiuguohua (approved):

PR #4399, opened 2025-06-03
Fix compile error & event dispath for spine cache mode

bofeng-song:


star-e (approved):

star-e:
@cocos-robot run test cases

